PR #24, opened 2018-10-04
More user-friendly installation page

OrdinaryJellyfish:
I made the primary installation page contain some instructions for basic installation. I also removed the links to the other OS guides (but didn't actually delete them, just in case) since @luceos didn't seem like he liked those (I agree, they would probably be better as tutorials in discuss). If you'd like me to add those, I can improve them (I already found an error in one :wink:) and add the links back.


datitisev:
@OrdinaryJellyfish Could you add those links back, and move the installation instructions to another file (`general.md`) and link it in `index.md`?

The OS guides are helpful, even if they are a bit wrong right now, to newbies. If you want, you can make a separate PR fixing issues with those specific instructions. People don't often check Discuss for installation instructions, only to get help.

OrdinaryJellyfish:
All right, I'll go ahead and do that.

OrdinaryJellyfish:
By the way, I'm not very good at CentOS so I won't be able to fix that article, just the Ubuntu one. I'll go ahead and make a PR for that page.

datitisev:
Committed as c772c8098cdbb1eec0225e832f45ad2e369f9477.

PR #26, opened 2018-10-15
Fix outdated extension filename

OrdinaryJellyfish:
Since beta.8 is introducing `extend.php` rather than `bootstrap.php` for extensions, the documentation needs that to be changed. I changed `boostrap` (typo intended 😁) to `extend`.


datitisev:
Thanks!

PR #27, opened 2018-10-15
Fix extend frontend class

OrdinaryJellyfish:
In `quick-start.md`, one example showed `Extend\Assets` instead of `Extend\Frontend`, and this PR fixes that.


CDK2020:
Missed that one, thanks!

OrdinaryJellyfish:
Perks of Ctrl+F :grin:

PR #28, opened 2018-10-22
Add locale registration warning

OrdinaryJellyfish:
This makes for a nice little warning for users who didn't register their locale first in their `extend.php` file.

### DIFF
--- a/extend/internationalization.md
+++ b/extend/internationalization.md
@@ -206,6 +206,10 @@ Support for grammatical gender will be added in a future version of Flarum. Deta
 
 # Server-side Translation
 
+::: WARNING
+In order for server-side translation to work, you must have your locale registered before everything else in your `extend.php`.
+:::
+
 Translation is generally handled by Flarum's front-end client. However, you can use translations in your PHP code if you need to.
 
 First, you'll need to get the `Flarum\Locale\Translator` class. This can be done a few different ways:
--- a/extend/internationalization.md
+++ b/extend/internationalization.md
@@ -206,10 +206,6 @@ Support for grammatical gender will be added in a future version of Flarum. Deta
 
 # Server-side Translation
 
-::: WARNING
-In order for server-side translation to work, you must have your locale registered before everything else in your `extend.php`.
-:::
-
 Translation is generally handled by Flarum's front-end client. However, you can use translations in your PHP code if you need to.
 
 First, you'll need to get the `Flarum\Locale\Translator` class. This can be done a few different ways:
@@ -224,6 +220,10 @@ You can learn more about the Translator's methods in [Symfony's `Translator` doc
 
 
 # Registering Locales
+
+::: WARNING
+In order for server-side translation to work, you must have your locale registered before everything else in your `extend.php`.
+:::
 
 There's one last thing you need to do before Flarum can use your translations. You need to register them. Fortunately, Flarum beta 8 makes this pretty easy.
 


datitisev:
Could you move right below the "Registering Locales" heading?

tobscure:
Out of curiosity what is a use-case where this is true?

OrdinaryJellyfish:
In one of my extensions, I registered the locale last, and whenever I tried
using translation server side, it sent the translation key rather than the
value, pushing the registration to the top fixed it.

datitisev:
Actually, I've never had this issue. ReFlar's extension generator adds locale last in the beta 8 version, and that's what I've used for [fof/custom-footer](https://github.com/FriendsOfFlarum/custom-footer/blob/master/extend.php). Locales are registered last and work just fine...

tobscure:
It would depend in what context you're using the translator, but with an extension fully converted over to the new Extenders API it shouldn't be a problem. So I think this change should be able to be reverted once we have extenders for everything.

luceos:
Aside from being unsure whether this should be in the docs due to Toby's reasons, I personally don't think that warnings/alerts should be beneath headings. I personally think they should go below the section explaining the content the alert belongs to.

Reasoning:

- An alert as first content gives off that it is complex, error-prone.
- An alert indicating you have to be careful of something you explain below is makes it less welcoming to read the actual content as the alert attracts most focus.
- An alert seems not intuitive, you're essentially saying "hey a new section about this warning and below the warning a longer description of the warning"

franzliedke:
Agreed that this can be considered a bug. I am *assuming* it might have to do with the problems mentioned in flarum/core#1578. I will add a note to our beta.9 project to revert this change.

@OrdinaryJellyfish Could you please provide the failing `extend.php` (with the "incorrect" order) so that we have something to test against?

OrdinaryJellyfish:
@franzliedke 

```php
<?php

namespace Reflar\EmailDigest;

use Flarum\Extend;
use Flarum\Forum\Content;
use Flarum\Foundation\Application;
use Illuminate\Contracts\Events\Dispatcher;

return [
    function (Application $app, Dispatcher $events) {
        $events->subscribe(Listeners\EnableEmailDigest::class);
        $app->register(Providers\SendDigestEmail::class);
    },
    (new Extend\Frontend('forum'))
        ->js(__DIR__ . '/js/dist/forum.js')
        ->route('/settings/email-digests', 'reflar_email_digest', Content\AssertRegistered::class),
    new Extend\Locales(__DIR__ . '/locale'),
];
```

I'm not exactly sure what happened, I was trying to send an email using translation keys with my `extend.php` layed out that way, and all it sent was the keys themselves rather than values. It may or may not have been a problem on my side, but the only thing that fixed it for me was moving it to the top.

datitisev:
@OrdinaryJellyfish For consistency anyway, you may want to always have the function at the bottom. The extenders are shortcuts and are easier to understand if they are at the top, instead of having to scroll through a possibly long function that uses the Dispatcher.

OrdinaryJellyfish:
The `EnableEmailDigest` listener doesn't use any translations, but here you go...
```php
<?php

namespace Reflar\EmailDigest\Listeners;

use Flarum\Event\ConfigureUserPreferences;
use Flarum\Extension\ExtensionManager;
use Flarum\User\AssertPermissionTrait;
use Illuminate\Contracts\Events\Dispatcher;

class EnableEmailDigest
{
    use AssertPermissionTrait;

    protected $ext;

    public function __construct(ExtensionManager $ext)
    {
        $this->ext = $ext;
    }

    public function subscribe(Dispatcher $events)
    {
        $events->listen(ConfigureUserPreferences::class, [$this, 'addUserPreference']);
    }

    public function addUserPreference(ConfigureUserPreferences $event)
    {
        $event->add('reflarEmailDigest', 'boolval', true);
        $event->add('reflar_emailDigest_timePeriod', 'strval', 'Daily');
        $event->add('reflar_emailDigest_contents_newDiscussions', 'boolval', true);
        $event->add('reflar_emailDigest_contents_newPosts', 'boolval', true);
        if ($this->ext->isEnabled('flarum-likes')) {
            $event->add('reflar_emailDigest_contents_likes', 'boolval', true);
        }
        if ($this->ext->isEnabled('reflar-reactions')) {
            $event->add('reflar_emailDigest_contents_reactions', 'boolval', true);
        }
        if ($this->ext->isEnabled('flarum-mentions')) {
            $event->add('reflar_emailDigest_contents_recentMentions', 'boolval', true);
        }
    }
}
```

I'd also prefer not to share any more in public, as this is going to be a premium extension. I'd gladly share the rest in a private means of communication though :wink:

franzliedke:
@OrdinaryJellyfish Not sure if this is still relevant, but I would suggest [replacing that event subscriber with a listener](https://github.com/flarum/subscriptions/commit/7ac73b834023e997147d4dd9c851a2ea73deba4b). Right now, you are always resolving the `ExtensionManager` (and all its dependencies) from the service container, whether the event is fired or not. That problem does not exist with listeners.

Also, if you don't mind, I'd love to see the service provider's source code - could you share it with me privately on Discord, for example?

franzliedke:
As we believe the underlying issue has been fixed in core and most extensions, we [removed](https://github.com/flarum/docs/commit/afc14e9f7a42539801c2eea1399fb0caf05c2ee4) the warning again. Thanks for your help!

PR #29, opened 2018-10-31
Update quick-start.md

luceos:
Take into consideration the `site-custom` initializer name when using extend.php from the root path. This is caused by not having an extension when including the JS, see:

https://github.com/flarum/core/blob/c446c5cc6174e5fde0d01495d80af126eb79804e/src/Extend/Frontend.php#L129-L132

### DIFF
--- a/extend/quick-start.md
+++ b/extend/quick-start.md
@@ -446,6 +446,8 @@ To figure out your initializer name use this table:
 | vendor/flarum-package     | vendor-package             |
 | vendor/package            | vendor-package             |
 
+> When using an `extend.php` inside the Flarum root directory, the initializer has to be named `site-custom`.
+
 OK, time to fire up the transpiler. Run the following commands in the `js/forum` directory:
 
 ```bash


tobscure:
This is actually not necessary, initialisers can be named anything and it doesn't matter. Also they are deprecated and will be replaced by JS extenders in b9 :D

luceos:
@tobscure using the extender, the JS initializer is added using `site-custom`, not naming your initializer will prevent the initializer from being available in the frontend. See:

https://github.com/flarum/core/blob/c446c5cc6174e5fde0d01495d80af126eb79804e/src/Frontend/Asset/ExtensionAssets.php#L59

Unless I'm totally off, I think properly naming the initializer inside the extend.php helped fix not seeing the code enacted in the frontend.

tobscure:
I think you're misunderstanding initializers. `app.initializers` is simply an ItemList. You can give your initializer a name (any name) when you add it so that other intializers can be added before or after it. Then the ItemList is converted to an array and the initializers are executed in order.

`flarum.extensions[name]` is a completed separate concept, that's an object that collects the exports of each extension's JS (currently no extensions export anything, that'll be the crux of the new JS extension API coming in b9 though).

luceos:
🤔 so what was previously the issue when using an initializer named differently than the shortened name of your extension? Eg using initializer key `flagrow-flarum-ext-upload` instead of `flagrow-upload`? There used to be an issue with that too.

Thanks for clarifying I'll see to getting a better grip of that part of core asap.

tobscure:
I don't know how there could've been an issue

PR #32, opened 2018-11-19
add shared hosting/subdirectory instructions

### LINKED ISSUE #15: New section "Security hardening"
As proposed by @Kulga [here](https://github.com/flarum/flarum/commit/569e3a3b5ec862ecae93e0e438dd6a7a4c5012ed#commitcomment-21080215), we should probably change our included Apache configuration (and the one for Nginx in the docs) to only include the bare essentials - or at least only things that cause no problems with other components.

(In this particular example, always stripping the `Proxy` request header might cause problems with reverse proxies.)

Instead, a section called "Security hardening" or something similar should contain further "recipes" with things that should be applied to secure the server configuration. (For things like the above, notes should make it clear that they should not be combined with certain setups, e.g. reverse proxies.)

### LINKED ISSUE #16: New section: "Shared hosts"
The default skeleton now requires pointing the webserver root to the public/ folder. This is not possible on most shared hosts.

For them, we need some additional instructions, they will mostly consist of:
- adding/moving one file (ping me for the exact code - I will do my best to keep the changes as small as possible)
- some necessary security-hardening (preventing access to certain files and directories, see flarum/core#692)

luceos:
fixes #15, fixes #16 

- [ ] decide on the location to add this in the navigation
- [ ] add to related pages, eg installation.md

### DIFF
--- a/user/custom-paths.md
+++ b/user/custom-paths.md
@@ -1,0 +1,57 @@
+# Custom paths
+
+These directions are meant for subdirectory installations or shared hosting where you have less control over your directory structure. For Flarum to work on these environments we need to modify the location of some of the files shipped with the software.
+
+## Public path
+
+Flarum serves the forum from a file called `index.php` this file makes sure requests for your forum, the underlying API and the admin are handled by the correct code.
+
+By default the `index.php` is located inside a folder named `public`. The idea is that this folder is the only folder made available when visitors browser your forum. That's also the reason why the `public/assets` directory contains your user avatars as these need to be visible by your users.
+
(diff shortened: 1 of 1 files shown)


franzliedke:
The paths also need to be adapted in the `flarum` CLI script as well.

tobscure:
Thanks @luceos. I will take care of merging this into my impending docs rewrite.

luceos:
@franzliedke I keep forgetting about that, good point!

@tobscure great, I hope this draft helped!

PR #33, opened 2018-11-22
Add Beta 7.2 url rewrite instructions link

jordanjay29:
To be used for the moment because Beta 8 is not yet out, and the written instruction set is for the future version.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -35,6 +35,10 @@ When everything is ready, navigate to your forum in a web browser and follow the
 
 ## URL Rewriting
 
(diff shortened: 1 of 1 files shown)


CDK2020:
I think you can simplify the first sentence. We should just tell people that the docs are for beta 8, no need to check if they are installing first.

tobscure:
We may as well just revert the instructions for now 0da3d157f92053d8f7cc8ea7f27968f92462d7c4

and merge https://github.com/flarum/docs/pull/34 when beta 8 is released

PR #42, opened 2018-11-30
clarification when moving index.php to root path

luceos:


franzliedke:
We should probably update the skeleton so that the base/public paths only need to be changed in a single location.

franzliedke:
That, and more: https://github.com/flarum/core/issues/1669

PR #43, opened 2018-12-03
Adding MariaDB minimum reqs

jordanjay29:
According to information from https://mariadb.com/kb/en/library/full-text-index-overview/


franzliedke:
[This installer code](https://github.com/flarum/core/blob/1106a2e3a33ca169232f73983a78b204e7231fb7/src/Install/Console/InstallCommand.php#L245-L247) should be adapted (and *fixed*), too.

PR #45, opened 2018-12-08
Update about shared hostings

littlefluffycat:


jordanjay29:
Closing this because it is not a supported installation method. 

To explain further, composer is used because Flarum has multiple dependencies and requires a minimum server environment which composer checks for before installation. Installing on a local environment (which may pass these checks) and then moving it to a new environment may mean the new server is not capable of running Flarum as is. There's also a large issue with a composer-less environment being unable to use extensions easily, they are possible only with a lot of overhead work. 

This is not an optimal solution and has created problems for users who follow it. We won't be adding instructions to the docs that will cause more confusion and complexity, the best real solution to being unable to install Flarum on a shared hosting environment without access to SSH is to contact the host directly for assistance or switch hosting providers.

CDK2020:
Hey @ghost, thanks for the PR! 

Unfortunately, shared hosting isn't technically supported by Flarum, and thus, I don't think this is something we should put on our documentation. Feel free to come chat with us in the docs channel of the Flarum Discord if you have any questions or concerns!

PR #47, opened 2018-12-11
Update contributing.md for macOS sed

liamwli:
The sed command in macOS requires a value to be passed to the -i flag. Using an empty string mimics the behaviour of other platforms.

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -26,7 +26,7 @@ git clone https://github.com/<username>/core.git packages/core
 git clone https://github.com/<username>/tags.git packages/tags # etc
 
 # Replace "^0.1.0" with "^0.1.x-dev" in composer.json
-sed -i 's/\^0\.1\.0/\^0\.1\.x-dev/' composer.json
+sed -i '' 's/\^0\.1\.0/\^0\.1\.x-dev/' composer.json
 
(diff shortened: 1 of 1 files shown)


tobscure:
Can someone please confirm that this does not break the command on linux?

luceos:
Confirmed the command in the docs is not working on OSX. Also confirmed that the new command does **NOT** work on Ubuntu.

franzliedke:
Confirmed *not* working on Ubuntu.

See [this question and answer](https://stackoverflow.com/questions/5694228/sed-in-place-flag-that-works-both-on-mac-bsd-and-linux) for even more fun. ;)

franzliedke:
How about just replacing this command with a manual step of changing `"minimum-stability": "beta"` to `"minimum-stability": "dev"` in composer.json? That's what I did in my local setup, and it does the job perfectly.

liamwli:
Yeah, it should be noted that even with this change, it didn't work anyway. I had to remove the carat to get composer to install the project with the cloned repos.

franzliedke:
I've [taken care of it](https://github.com/flarum/docs/commit/dce4702c07c810f42a05e533c973da06b8d0289c).

PR #49, opened 2018-12-19
Change 0.1.0-beta.8 version constraint

OrdinaryJellyfish:
Looks like all those extensions people are having problems installing with beta 8.1 is due to this little missing ^ in the getting started page :)

### DIFF
--- a/docs/extend/start.md
+++ b/docs/extend/start.md
@@ -89,7 +89,7 @@ We need to tell Composer a bit about our package, and we can do this by creating
     "description": "Say hello to the world!",
     "type": "flarum-extension",
     "require": {
-        "flarum/core": "^0.1.0-beta.8"
+        "flarum/core": "0.1.0-beta.8.*"
     },
     "autoload": {
         "psr-4": {"Acme\\HelloWorld\\": "src/"}
--- a/docs/extend/start.md
+++ b/docs/extend/start.md
@@ -89,7 +89,7 @@ We need to tell Composer a bit about our package, and we can do this by creating
     "description": "Say hello to the world!",
     "type": "flarum-extension",
     "require": {
-        "flarum/core": "0.1.0-beta.8.*"
+        "flarum/core": ">=0.1.0-beta.8 <0.1.0-beta.9"
     },
     "autoload": {
         "psr-4": {"Acme\\HelloWorld\\": "src/"}
--- a/docs/extend/start.md
+++ b/docs/extend/start.md
@@ -89,7 +89,7 @@ We need to tell Composer a bit about our package, and we can do this by creating
     "description": "Say hello to the world!",
     "type": "flarum-extension",
     "require": {
-        "flarum/core": "0.1.0-beta.8"
+        "flarum/core": "^0.1.0-beta.8"
     },
     "autoload": {
         "psr-4": {"Acme\\HelloWorld\\": "src/"}


tobscure:
This was actually intentional because the `^` prefix will mean the extension says it's compatible with future beta versions (but there may be breaking changes). But you're right that being specific for beta.8 is problematic. Perhaps would `0.1.0-beta.8.x` work?

franzliedke:
Yeah. Does `0.1.0-beta.8.*` (star, not `x`) work for you, @OrdinaryJellyfish?

Magicalex:
It's doesn't work 

`Skipped branch master, Could not parse version constraint 0.1.0-beta.8.*: Invalid version string "0.1.0-beta.8.*"`

tobscure:
@franzliedke I thought the correct syntax for this was `.x`, not `.*` ?

luceos:
If I remember correctly ;

Wildcards don't work with stability strings,
x only works with branches or branch aliases,
I think we best suggest exact constraints until we reach stable or accept a lower constraint is not precise.

tobscure:
https://getcomposer.org/doc/articles/versions.md#wildcard-version-range-
yep I think you are right @luceos

jordanjay29:
> I think we best suggest exact constraints until we reach stable

This would be easiest for support. IMHO, it'd be better to see select community members stepping up to provide compatible version forks rather than have the larger community confused on what does/doesn't work on their Flarum site as far as extensions go. 

It also incentivizes extension devs to stay apprised of Flarum releases and perhaps work on features on their extension's to-do list.

rob006:
> I think we best suggest exact constraints until we reach stable

This is a terrible idea. If each extension will use `0.1.0-beta.8` as a constraint, I will not be able to update to  `0.1.0-beta.8.1` until each of them creates a new release with updated constraint. This may block security updates.

As a continuation of the discussion from [forum](https://discuss.flarum.org/d/17745-flarum-0-1-0-beta-8-released/169), I don't think that this is possible without changing `flarum/core` versioning policy - right now it is extremely impractical. 

Using standard semver versioning for pre-1.0 releases would look like:
* BC breaking releases as `0.X.0`, 
* BC compatible feature releases as `0.x.Y`,
* patch/security releases as `0.x.y.Z`.

Extensions may then just use `^0.1` as a constraint.

franzliedke:
@rob006 I think the main question here is when to release a version that we want to call 0.x (first version without a "beta" label - what we want to consider "feature complete" and "ready for production"). We haven't yet felt ready to flip that switch, but are actively working towards doing so as soon as possible.

rob006:
> I think the main question here is when to release a version that we want to call 0.x (first version without a "beta" label - what we want to consider "feature complete" and "ready for production")

I think that something is fundamentally wrong with this question: "feature complete" and "ready for production" should be released as `1.0.0`, not `0.1.0`. Don't be afraid to release stable and production-ready software as `1.0.0`, the whole pre-1.0 phase is for experiments and BC breaks. It would be much simpler if we would be at `0.8.1-beta` stage than `0.1.0-beta.8.1`.

@jordanjay29 `beta` suffix is not an excuse for every weird thing -  current versioning is confusing and impractical, and this issue is the best example of it.

> Yes, being more strict on versioning may be an inconvenience for some users, but it also provides a safety net for users in general

Plugins/libraries should not use too strict constraints - using exact version as constraint is a very bad practice. The whole problem was solved ages ago with semver, Flarum just needs to start following it.

CDK2020:
We could just do `>=0.1.0-beta.8 <0.1.0-beta.9`

franzliedke:
In hindsight, adopting a 0.x versioning scheme earlier would have been smarter. Right now, it's not (yet) the right signal.

How do we want to improve this problematic documentation, though? Personally, I'd even say a constraint promising compatibility with every (beta?) version greater than the current one is fine. Everything is better than extensions blocking core updates.

franzliedke:
I have finally written up my proposal for how to solve this problem - and avoid having extensions prevent core upgrades - over in #75.

Hence, I am closing this.

@OrdinaryJellyfish @rob006 @CDK2020 Thanks for the lively discussion! I would love to have your input over on the other PR. This is an important topic. :heart:

PR #50, opened 2018-12-27
Update Installation Docs

frinkfromthedrink:
This PR adds the `php-curl` extension as it is required in the install. Not sure if it is a Flarum dependency, or a dependency of something that Flarum depends on....

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -11,7 +11,7 @@ Beta is all about fixing these issues and improving Flarum. **Please don't use F
 Before you install Flarum, it's important to check that your server meets the requirements. To run Flarum, you will need:
 
 * **Apache** (with mod_rewrite enabled) or **Nginx**
-* **PHP 7.1+** with the following extensions: dom, gd, json, mbstring, openssl, pdo_mysql, tokenizer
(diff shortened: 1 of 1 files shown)


franzliedke:
I think we've had this discussion before. It's not really a Flarum dependency, but one of Composer (which you need to install first before you can install Flarum). The list in our documentation should match that enforced in our installer, which doesn't check curl, because it's already too late at that point.

franzliedke:
See https://github.com/flarum/flarum.github.io/pull/69. Please find out what dependency is requiring curl. Paste the full output and we might be able to help you. :wink:

frinkfromthedrink:
I can't see any curl requirements in any of the flarum extensions, or the core code....

franzliedke:
@frinkfromthedrink That's not what I meant. How did you encounter the problem in the first place? Through some error about curl missing, I suppose. We need to find the source of that error. If it's Composer, then IMO it is not our dependency.

frinkfromthedrink:
I believe it is a composer dependency, @jordanjay29 and I had a discussion about it, and I think composer uses curl to download the files, so this can probably be closed then

PR #51, opened 2019-01-14
Update start.md

ftchao:


docs/extend/start.md
```diff
@@ -145,4 +145,4 @@ Once that's done, go ahead and fire 'er up on your forum's Administration page,
 
 Woop! Hello to you too, extension!
 
-We're making good progress. We've learned how to set up our extension and use extenders, which opens up a lot of doors. Read on to learn how to extend Flarum's frontend.
\ No newline at end of file
+We're making good progress. We've learned how to set up our extension and use extenders, which opens up a lot of doors. Read on to learn how to extend Flarum's frontend.
```
luceos: What is the change in this line?

datitisev:
I don't see why this change is required. One can use `composer require <package> <version>` to install packages.
From the composer help:

> **packages**:
> Optional package name can also include a version constraint, e.g. foo/bar or foo/bar:1.0.0 or foo/bar=1.0.0 or "foo/bar 1.0.0"

The only issue I see with the old approach that uses `*@dev` is that some shells may try to parse `*` into file names on the current folder, which can be fixed by wrapping it with quotes.

franzliedke:
No title, no description. I am closing this.

luceos:
@datitisev the quote you posted shows exactly what the issue is, when you use a version where you separate it with a space you need to use quotes!

Re-opening for that reason, the current docs are incorrect. I'm just unsure why that one line was modified.

datitisev:
@luceos I think they only included the quotes to show that that's the argument and that the two go together. Either way, I've always used `composer require <package> <version>` with no issues, it has always added the `<version>` to composer.json. 

See this following screenshot. Uses `composer require <package1> <ver1> <package2> <ver2>`.

image

luceos:
@datitisev that's a first, because in the past this did not work, no kiddin.

Then okay, fine to close 😂

PR #69, opened 2019-08-22
Beta 10 Upgrade Guide

franzliedke:
A first, preliminary version of a Beta 10 Upgrade Guide for extension developers.

We can collect breaking changes and other relevant items here and then publish it around the time we go into feature freeze mode (to give extension developers a little time to prepare their extensions for the new version, if necessary).

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -33,6 +33,7 @@ module.exports = {
             '/extend/data',
             '/extend/distribution',
             '/extend/update-b8',
+            '/extend/update-b10',
           ]
         },
         {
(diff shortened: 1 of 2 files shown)


CDK2020:
Lifecycle extenders are new as well so we should include those.

Priority should also be added pending https://github.com/flarum/core/pull/1838

franzliedke:
> Lifecycle extenders are new as well

Not in Beta 10, though. :thinking: 

> Priority should also be added pending flarum/core#1838

Still thinking about that one, to be honest. Priority numbers feel a little too brittle.

luceos:
let's make sure it mentions being work in progress for items being added, then we can publish this

luceos:
@datitisev mentioned it might be wise to give more attention to breaking changes, so move it up?

docs/.vuepress/config.js
```diff
@@ -33,6 +33,7 @@ module.exports = {
             '/extend/data',
             '/extend/distribution',
             '/extend/update-b8',
+            '/extend/update-b10',
```
franzliedke: Should we remove the old upgrade guide for beta.8? At least from the menu?

datitisev: @franzliedke I'd leave it be, could be useful for old extensions that developers want to revive.

PR #70, opened 2019-08-23
Upgrade to Vuepress v1 and use GitHub Actions

datitisev:
Upgrading to v1 simply fixed some style glitches, like the external link button always being too high.

image


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,29 @@
...
+    - name: Install Dependencies
+      run: npm ci
+
+    - name: Build
+      run: npm run build
+      if: github.event.deleted == false
+
+    - name: Deploy
+      uses: peaceiris/actions-gh-pages@v1.1.0
+      if: success()
```
franzliedke: Isn't that the default anyway?

datitisev: I think so, but the example code had that if statement and didn't set `continue-on-error` to true, so... 🤷‍♂

franzliedke: Yep, it's necessary, because the previous step is executed conditionally, and if it doesn't execute, its status is "neutral".

Full explanation: https://github.com/peaceiris/actions-gh-pages/commit/6492b522cec8813590f3f31094b75c43725adccf#commitcomment-34950365

luceos:
Docs are published via the flarum.org website now. It is no longer using the github pages implementation 😊

franzliedke:
Okay, I removed the GH action again.

Some better automation (daily builds?) would be nice, but I guess it's not particularly urgent. I'd like to see the docs better integrated into the site layout as well.

PR #71, opened 2019-08-25
Fixes Caddy Configuration

tankerkiller125:
Removed random comma that doesn't belong

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -70,7 +70,7 @@ www.example.com {
     fastcgi / /var/run/php/php7.2-fpm.sock php
     header /assets {
         +Cache-Control "public, must-revalidate, proxy-revalidate"
-        +Cache-Control "max-age=25000",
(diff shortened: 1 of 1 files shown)


franzliedke:
Thanks a lot! :heart: Are you using Caddy?

tankerkiller125:
Yes, I am using Caddy to host my forum and in fact all of my websites.

franzliedke:
Cool, haven't come across too many people using it yet. Seems very slick!

tankerkiller125:
The automatic LetsEncrypt implementation is why I originally started using it for personal sites, It's sense grown on me and some of the features are just too good to pass up on. Notably when on larger infrastructure the ability for it to get the certs and then automatically share it across every instance (Auto Scaling Anyone?) using redis has just made it the default tool in my toolbox, I only use Nginx and Apache when I absolutely have too and even then I usually proxy it through caddy.

PR #72, opened 2019-08-25
Fix Component import not working

datitisev:
Not sure why we mix `@flarum/core/forum` and `flarum/`, but only the latter worked for someone on Discord, so we should probably change it anyway.

### DIFF
--- a/docs/extend/frontend.md
+++ b/docs/extend/frontend.md
@@ -173,7 +173,7 @@ You should familiarize yourself with [Mithril's component API](https://mithril.j
 With these differences in mind, a basic component might look like this:
 
 ```jsx
-import { Component } from '@flarum/core/forum';
+import Component from 'flarum/Component';
 
 class Counter extends Component {
   init() {


franzliedke:
The "Not sure why" part always makes me nervous. :wink:

Would `import { Component } from '@flarum/core/common';` work as well?

datitisev:
@franzliedke No, as `webpack.config.js` only adds `@flarum/core/forum` and `@flarum/core/admin` as aliases, I think. We should stick to one format regardless, this can be changed in the frontend rewrite, and I don't think `@flarum/core` is used in any extension.

(not saying to remove the support for it, just that there's not really a reason to use it right now)

tobyzerner:
`@flarum/core` refers to https://github.com/flarum/core/blob/master/js/package.json
`@flarum/core/forum` refers to https://github.com/flarum/core/blob/master/js/forum.js

Public API should be explicitly exported from
https://github.com/flarum/core/blob/master/js/src/common/index.js and 
https://github.com/flarum/core/blob/master/js/src/forum/index.js#L13-L15

To prevent extensions from actually importing and compiling with the `@flarum/core` code in their dist file, `flarum-webpack-config` exposes `@flarum/core/forum` as an [external](https://github.com/flarum/flarum-webpack-config/blob/master/index.js#L55) (all of the exports from `@flarum/core/forum` are [assigned](https://github.com/flarum/core/blob/master/js/webpack.config.js#L7) to `window.flarum.core`).

Therefore:

```js
import { Component } from '@flarum/core/forum';
```

is effectively translated to the following by webpack:

```js
const Component = flarum.core.Component;
```

Same deal for extensions; exports are assigned to `window.flarum.extensions['vendor/package']`. To use it, you add [`useExtensions`](https://github.com/flarum/flarum-webpack-config#useextensions) to your webpack config to set up the import redirect from `@vendor/package/forum` to `window.flarum.extensions['vendor/package']`.

The advantage of setting it up this way? Because you're importing stuff from a real package, you can get IDE autocompletion.

---

Importing from `flarum/*` is legacy. `@flarum/core/forum` exports a `compat` object (see [here](https://github.com/flarum/core/blob/master/js/src/forum/index.js#L17-L22) and [here](https://github.com/flarum/core/blob/master/js/src/forum/compat.js)). `flarum-webpack-config` directs any `flarum/*` imports to [read from that compat object](https://github.com/flarum/flarum-webpack-config/blob/master/index.js#L74-L81).

datitisev:
@tobyzerner Not sure if typo, but Flarum exports go to `flarum.core.compat`, not `flarum.core`

Either way, I don't really like this method as it requires extensions to export classes for other extensions to use rather than having them readily available. We can discuss this on the frontend framework rewrite though.

I don't really see the reason as to why `flarum/*` is legacy (and if it is, why are we using it in all of our official extensions ??). Both `@flarum/core/forum/` and `flarum/` point to the exact same thing, so there's really no reason to include the `forum` or `admin` specification, as it means nothing anyway...

PR #75, opened 2019-09-04
Extensions: Tweak recommended version constraints

franzliedke:
Here's my proposal for reducing the compatibility problems with extensions, as a follow-up to the discussion in #49.

The idea: let extensions declare they're compatible with the current and upcoming versions of Flarum.

Benefits:
- Extensions following this recommendation can no longer prevent upgrades to the next version of Flarum.
- Extension authors have more time to make extensions compatible with new APIs and features in core.

Drawbacks:
- In theory, it is possible to install extensions that aren't compatible with your version of Flarum.

The proposed deprecation cycle of two months is meant to reduce the chance of that drawback becoming a problem, and give extension authors time to update their extensions without a rush after a new release has appeared. So we would have to be careful not to introduce any breaking changes whenever possible, unless we have offered (and advertised!) the alternative API for at least one beta release cycle.

---

NB: In hindsight, our versioning scheme doesn't work so well anymore. One of the reasons why our primary goal is to make the final breaking changes to our extension API so that we can finally declare 0.1 is done. Until that is the case, we have to adapt as best as we can.

---

**Things to review:**

- Wording okay?
- Is the explanation understandable? Is it too long?

### DIFF
--- a/docs/extend/start.md
+++ b/docs/extend/start.md
@@ -89,7 +89,7 @@ We need to tell Composer a bit about our package, and we can do this by creating
     "description": "Say hello to the world!",
     "type": "flarum-extension",
     "require": {
-        "flarum/core": "0.1.0-beta.8"
+        "flarum/core": ">=0.1.0-beta.10 <0.1.0-beta.12"
     },
     "autoload": {
         "psr-4": {"Acme\\HelloWorld\\": "src/"}
@@ -117,7 +117,18 @@ We need to tell Composer a bit about our package, and we can do this by creating
 
 * **require** contains a list of your extension's own dependencies.
   * You'll want to specify the version of Flarum that your extension is compatible with here.
-  * Since Flarum is still in beta, do **not** specify a version range or prefix for the Flarum version, otherwise your extension may become installable on future incompatible beta versions.
+  * This is also the place to list other Composer libraries your code needs to work.
+
+  ::: warning Carefully choose the Flarum version
+  While Flarum is still in beta, we recommend that you declare compatibility both with the current and the upcoming beta version of Flarum:
+
+      "flarum/core": ">=0.1.0-beta.10 <0.1.0-beta.12"
+
+  This gives you time to update your extension for new features or changes in Flarum's core, without preventing users from upgrading to the latest Flarum release because your extension is not compatible.
+
+  To make this possible, we try to deprecate features for one beta cycle, before removing them for the next one, until we reach stable.
+  This gives you two months time to update.
+  :::
 
 * **autoload** tells Composer where to find your extension's classes. The namespace in here should reflect your extensions' vendor and package name in CamelCase.
 


datitisev:
> Extensions following this recommendation can no longer prevent upgrades to the next version of Flarum.

Upgrades would be prevented for beta 12 & later, though I assume u mean beta 10 & beta 11, including bugfix updates.

And extensions should still largely work on beta 9 as well, if they don't use `ForbiddenException`...

I'm not really sure. There's not a good solution for this problem unless we use `MAJOR.MINOR.BUGFIX` versioning (as `^1.0.0` would include new features and no breaking changes, but `^0.1.0` would not include new features, just bug fixes... unless we don't follow semver and use `0.MAJOR.MINOR` instead).

The software _is_ in beta, after all, so I don't really see an issue with some users experiencing issues with outdated third-party extensions... and not allowing the use of extensions that work perfectly fine but have been abandoned is probably not a good idea either.

---

Aside from that, if we do stick to this PR, it looks fine to me. Perhaps add that users should make sure to update their extension even if it would still work, as the constraint would not allow installation ?

franzliedke:
> Upgrades would be prevented for beta 12 & later, though I assume u mean beta 10 & beta 11, including bugfix updates.

That's the point: we assume that extensions will normally still work (or at least the rest of the forum will) with the next beta release as well. At the same time, we don't want "very old" extensions (defined to be more than two beta cycles old) to be installable on newer betas. That was the motivation behind the previous recommendation, albeit that was even stricter.

(10, 11 and 12 are just the currently relevant examples.)

rob006:
> At the same time, we don't want "very old" extensions (defined to be more than two beta cycles old) to be installable on newer betas.

It may be too late for that for existing ecosystem. Most of extensions already has releases with `^0.1.0-beta.8` as constraint. Even if they start using `>=0.1.0-beta.10 <0.1.0-beta.12` in future releases, old releases will be always installable on any `0.1.x` release of Flarum. Using more strict versioning for future releases may be even worse in some cases - even if your extensions with `>=0.1.0-beta.10 <0.1.0-beta.12` as a constraint works fine on `0.1.0-beta.12`, Composer will pick older release, which has zero chance to work on this version.

Also note that it is not only "beta problem". After releasing stable `0.1.0` of Flarum, everything will stay the same - old beta 7 extensions still will be installable by Composer.

franzliedke:
@rob006 Surely that problem can always be solved by releasing a new version and asking consumers to use that version of the extensions.

Authors have to take care of sensible constraints, of course - this is just trying to improve our default recommendation for the current phase.

In general, I would like to move towards separating the set of dependencies we have for core vs. those for extensions, but that is a completely different topic and requires a lot of Composer hacking.

rob006:
@franzliedke I'm just pointing out that there are issue with this approach. Issues that would not exist if you use more standard semver and release new BC breaking release as `0.2.0` (still as beta - pre-1.0.0 releases are exactly for such cases as Flarum current state).

franzliedke:
Thanks. We've discussed this approach internally and decided we don't want to change the versioning strategy right now; because of all the communication issues that would entail.

Instead, we're pushing as hard as we can to reach our goals for the first stable release - I hope the last months' activity showed that we're on a good track.

CDK2020:
If we stick to the deprecation stance that is written in the changes, this seems fine to me

Issue #62, opened 2019-06-23
Create section to address file ownership during installation


jordanjay29:
Users are often presented with an error during installation to chmod their files to 775, and many, upon completing it, find the error still persisting and Flarum not installing. This often stems from improper ownership/authority of the file owner and not the permissions of the file itself. 

A section on the Installation page should be written to describe, in simple yet informative terms, what is occurring and what Flarum requires to proceed. In addition, we should include an example based on *commonly advised setups* (e.g. www-data as the php/webserver user) with a disclaimer that individual environments may vary.

jordanjay29:
Fixed with https://github.com/flarum/docs/pull/78

PR #78, opened 2019-09-30
Add permissions and ownership instructions

tankerkiller125:
This pull request addresses #62

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -77,7 +77,7 @@ www.example.com {
 }
 ```
 ## Folder Ownership
-During installation Flarum may request that you set some folders to permission level `755`. If after completing these steps Flarum continues to request that you change the permissions you may need to check that your files are owned by the correct group and user. 
+During installation, Flarum may request that you set some folders to permission level `755`. If after completing these steps, Flarum continues to request that you change the permissions you may need to check that your files are owned by the correct group and user. 
 
 By default in most linux distributions, `www-data` is the group and user that both PHP and the web server operate under. You can change the folder ownership in most linux operating systems by running `chown -R www-data:www-data foldername/`. 
 
--- a/docs/install.md
+++ b/docs/install.md
@@ -79,7 +79,7 @@ www.example.com {
 ## Folder Ownership
 During installation Flarum may request that you set some folders to permission level `755`. If after completing these steps Flarum continues to request that you change the permissions you may need to check that your files are owned by the correct group and user. 
 
-By default in most linux distributions `www-data` is the group and user that PHP and the web server operate under. You can change the folder ownership in most linux operating systems by running `chown -R www-data:www-data foldername/`. 
+By default in most linux distributions, `www-data` is the group and user that both PHP and the web server operate under. You can change the folder ownership in most linux operating systems by running `chown -R www-data:www-data foldername/`. 
 
 ::: warning Environments may vary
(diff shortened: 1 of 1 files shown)


docs/install.md
```diff
@@ -76,6 +76,18 @@ www.example.com {
...
+During installation Flarum may request that you set some folders to permission level `755`. If after completing these steps Flarum continues to request that you change the permissions you may need to check that your files are owned by the correct group and user. 
+
+By default in most linux distributions `www-data` is the group and user that PHP and the web server operate under. You can change the folder ownership in most linux operating systems by running `chown -R www-data:www-data foldername/`. 
+
+::: warning Environments may vary
+Your environment may vary from the documentation provided, please consult your web server configuration or web hosting provider for the proper user and group that PHP and the web server operate under.
+:::
+
+::: danger Never use permission 777
+You should never set any folder or file to permission level `777` this permission level allows anyone to access the content of the folder and file regardless of user or group.
```
franzliedke: `777, as this`? Or start a new sentence there?

franzliedke:
Generally, I like this. Leaving for @jordanjay29 to merge, though, as I remember seeing you discuss this topic in chat.

That said, we should also consider making the permission checks more intelligent in the installer - that would be even cooler than thorough documentation on this part. :grinning:

docs/install.md
tankerkiller125: Thanks for pointing that out, I read through this multiple times but usually at night so I probably was tired and missed it, I'll fix that tonight after work.

docs/install.md
```diff
@@ -76,6 +76,18 @@ www.example.com {
     gzip
 }
 ```
+## Folder Ownership
+During installation Flarum may request that you set some folders to permission level `755`. If after completing these steps Flarum continues to request that you change the permissions you may need to check that your files are owned by the correct group and user.
```
jordanjay29:
```suggestion
During installation, Flarum may request that you set some folders to permission level `755`. If after completing these steps, Flarum continues to request that you change the permissions you may need to check that your files are owned by the correct group and user. 
```

docs/install.md
```diff
@@ -76,6 +76,18 @@ www.example.com {
     gzip
 }
 ```
+## Folder Ownership
+During installation Flarum may request that you set some folders to permission level `755`. If after completing these steps Flarum continues to request that you change the permissions you may need to check that your files are owned by the correct group and user. 
+
+By default in most linux distributions `www-data` is the group and user that PHP and the web server operate under. You can change the folder ownership in most linux operating systems by running `chown -R www-data:www-data foldername/`.
```
jordanjay29:
```suggestion
By default in most linux distributions, `www-data` is the group and user that both PHP and the web server operate under. You can change the folder ownership in most linux operating systems by running `chown -R www-data:www-data foldername/`. 
```

docs/install.md
jordanjay29:
```suggestion
You should never set any folder or file to permission level `777`, as this permission level allows anyone to access the content of the folder and file regardless of user or group. 
```

This one seemed to get missed. Fix and I'll approve the PR.

tankerkiller125: @jordanjay29 The intent was to fix it, unfortunately I got very busy at work and had to do some work from home so I didn't have time to do it immediately.

PR #79, opened 2019-09-30
Fix code block spacing being inconsistent

tankerkiller125:
This fixes some code blocks having 4 spaces before the code and some other not having 4 spaces before hand.


franzliedke:
I am thinking the intention here was to make it easier to copy this with the correct indentation.

But since at least some of these snippets [will likely be obsolete soon](https://github.com/flarum/core/issues/1669), the whitespace isn't technically required and consistency is nice, too, I will merge. Thanks! :raised_hands:

franzliedke:
Whoops, wrong button.

tankerkiller125:
@franzliedke During testing at least Vuepress removed the indentation anyways. So I'm not entirely sure  if maybe what you said was the case or if someone just copied from the original code and then didn't remove the extra whitespace.

franzliedke:
I am confused. The idea was to have the extra spaces because they match the existing indentation in the code files that need editing.

But that's all moot now anyway, thanks for your help!

PR #80, opened 2019-10-03
Add Middleware Guide

OrdinaryJellyfish:


franzliedke:
I believe this new file needs to be added to `docs/.vuepress/config.js` to appear in the table of contents.

docs/extend/middleware.md
```diff
@@ -1 +1,108 @@
-#  Middleware
\ No newline at end of file
+#  Middleware
+
+Middleware is a nifty way to filter HTTP requests in Flarum. This can allow you to, for example, run a check against an external database checking a user when they register, either allowing them to continue the process or not. The possibilities are endless!
```
franzliedke: I would suggest the following:

> Middleware is a nifty way to wrap the handling of HTTP requests in Flarum.

And then add an example that clarifies that you can do something to the request *before* handling as well as something to the response *after* handling.

docs/extend/middleware.md
```diff
@@ -1 +1,108 @@
...
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+
+class YourMiddleware implements MiddlewareInterface {
+  public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+  {
+    // Your logic here!
+  }
+}
+```
```
franzliedke: I would suggest moving this middleware logic (with at least a `return $handler->handle($request)`) to the top, right below the intro paragraph. For somebody who does not know what middleware does, seeing one is probably more important than learning how to register them with Flarum.

docs/extend/middleware.md
```diff
@@ -1 +1,108 @@
...
+#  Middleware
+
+Middleware is a nifty way to filter HTTP requests in Flarum. This can allow you to, for example, run a check against an external database checking a user when they register, either allowing them to continue the process or not. The possibilities are endless!
+
+## Adding Middleware In Your Extension
+
+To add a new middleware, add a listener in your `extend.php` file:
+
+```php
+$events->subscribe(Listeners\AddMiddleware::class);
```
tankerkiller125: This will no longer be needed in beta.12 this can be changed to:

```
use Flarum\Extend

return [
    // Add middleware to user facing Forum
    (new Extend\Middleware('forum'))->add(MiddlewareName::class),
    // Add middleware to API
    (new Extend\Middleware('api'))->add(MiddlewareName::class),
    // Add middleware to admin frontend
    (new Extend\Middleware('admin'))->add(MiddlewareName::class),
];
```

OrdinaryJellyfish: Yeah, saw that. Great improvement. And you can make the middleware exclusive to specific frontends too?? Very nifty.

docs/extend/middleware.md
```diff
@@ -1 +1,108 @@
...
+```
+
+Now that we've got the basics down, let's run through a few more things:
+
+## Restricting Middleware to Certain Routes
+
+If you don't need your middleware to execute under every route, you can add an `if` to filter it:
+
+```php
+use Zend\Diactoros\Uri;
```
tankerkiller125: This will need to change to `use Laminas\Diactoros\Uri;` for beta.12 onward.

OrdinaryJellyfish:
@franzliedke `docs/.vuepress/config.js` already contains the page, as it already existed, just empty.

tankerkiller125:
@OrdinaryJellyfish It's just a blank page right now https://flarum.org/docs/extend/middleware.html

OrdinaryJellyfish:
@tankerkiller125 yeah, I'm guessing they were made as placeholders for guides to be written in the future. It won't be empty anymore as soon as this PR is approved & merged though :)

docs/extend/middleware.md
```diff
@@ -1 +1,109 @@
...
+```php
+use Laminas\Diactoros\Uri;
+
+$currentRoute = $request->getUri()->getPath();
+$routeToRunUnder = new Uri(app()->url('/path/to/run/under'));
+
+if ($currentRoute === $routeToRunUnder->getPath()) {
+  // Your logic here!
+}
+```
```
franzliedke: Keen to hear your (and other readers') opinion here, but I think this might be confusing without at least the skeleton of the `handle()` method. So, IMO, at least that function body should always be included. 

If we want to keep the example short, we can avoid the `use` lines by using fully-qualified class names everywhere. But that's a tradeoff...

docs/extend/middleware.md
```diff
@@ -1 +1,109 @@
...
+
+If you'd like to do something with the response after the initial request as been handled, that's no problem! Just run the request handler and then your logic:
+
+```php
+$response = $handler->handle($request);
+
+// Your logic here!
+
+return $response;
+```
```
franzliedke: Can we amend this with some text reminding the reader that PSR-7 responses are immutable, which means the `$response` variable will have to be re-assigned every time when calling a method such as `->withHeader()`?

franzliedke:
@luceos @jordanjay29 
Can we live with delaying the publication? Or do we want to publish some intermediate form, or include a warning about content referring to the upcoming beta release?

jordanjay29:
@franzliedke We can delay, Beta 12 isn't far away, and we could mention this issue in the Dev Blog (or even in #announcements on Discord) in the meantime to spread the word for extension devs.

franzliedke:
> we could mention this issue in the Dev Blog

@jordanjay29 Good idea... [done](https://discuss.flarum.org/d/22183-dev-diary-beta-12/28).

tankerkiller125:
This needs to be updated to include examples of `remove()`, `insertBefore()`, `insertAfter()` and `replace()` functions for the extender.

OrdinaryJellyfish:
@franzliedke changed to the CSRF token middleware, comments added to clarify

PR #82, opened 2019-10-18
Expand on mod_rewrite requirement

OrdinaryJellyfish:
Because of that long support request in #support 😛

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -25,6 +25,10 @@ At this stage, it's not possible to install Flarum by downloading a ZIP file and
 This doesn't necessarily mean you need a VPS. Some shared hosts give you SSH access, through which you should be able to install Composer and Flarum just fine. For other hosts without SSH, you can try workarounds such as [Pockethold](https://github.com/andreherberth/pockethold).
 :::
 
+::: warning mod_rewrite and Apache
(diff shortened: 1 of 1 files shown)


docs/install.md
```diff
@@ -41,14 +41,16 @@ When everything is ready, navigate to your forum in a web browser and follow the
 
 ### Apache
 
-Flarum includes a `.htaccess` file in the `public` directory – make sure it has been uploaded correctly. If you're using shared hosting, confirm with your provider that `mod_rewrite` is enabled and `.htaccess` files are allowed. If you're managing your own server, you may need to add the following to your site configuration:
+Flarum includes a `.htaccess` file in the `public` directory – make sure it has been uploaded correctly. **Flarum will not function properly if `mod_rewrite` is not enabled and `.htaccess` is not allowed.** Be sure to check with your hosting provider (or your VPS) that these features are enabled. If you're managing your own server, you may need to add the following to your site configuration to enable `.htaccess` files:
```
franzliedke: > `mod_rewrite` is not enabled and `.htaccess` is not allowed

IMO, that should be:

> `mod_rewrite` is not enabled or `.htaccess` is not allowed

That said, what about adding a line or two about enabling `.htaccess` overrides? If I remember correctly, this isn't enabled by default anymore in recent versions of Apache.

OrdinaryJellyfish: Done. How's it look now?

franzliedke: Nice, thank you!

PR #88, opened 2020-03-12
Update Symfony links to point to version 3.3

clarkwinkelmann:
The PluralizationRules link was broken because it no longer exists in the latest Symfony. The other links pointed to outdated or too recent documentation.


franzliedke:
In my local installation, I did not have a single Symfony package that's locked to 3.3.*. :thinking:

clarkwinkelmann:
I must have messed up something because I used one link for 3.4 and all others for 3.3 :thinking: I meant to link 3.3

The reason to link to 3.3 docs is because we require `^3.3` in Flarum. Consulting the documentation for 3.4 might show features that not everyone will have available.

clarkwinkelmann:
It might be logical to bump the requirement from 3.3 to 3.4 in Flarum, as 3.4 is the only supported release of 3.*. But in the meantime I still think linking 3.3 makes the most sense in the documentation.

franzliedke:
Fair enough. Feel free to bump our requirements to 3.4. :smiley:

PR #95, opened 2020-05-08
Start of appearance page

DoperBeats:


askvortsov1:
Good start, the Appearance page is definitely a better place for this documentation. Could you update it a bit to reflect all the customization options available on Appearance? 

Also, do we want to put instructions on accessing the admin dashboard somewhere centralized?

KyrneDev:
I think this is a nice change, but should be dependent on an actually "basic admin" page. We don't want to be telling people how to get to the ACP in every page about the ACP.

askvortsov1:
@KyrneDev another question is, considering that the admin dashboard is somewhat self explanatory, do we need explanation of how to use it?

luceos:
I second your opinion @askvortsov1 ; if the admin page isn't clear enough let's improve the admin page not divert a users' attention to the docs. That sounds like an unreasonable waste of time for that user.

Closing for now, unless someone raises objections.

PR #101, opened 2020-06-15
Update FAQ, Remove old links/outdated info

tankerkiller125:
Removes outdated links/information, adds Github Sponsors link, removes references to the outdated Roadmap.


franzliedke:
That roadmap is still the one that comes closest to a source of truth.

PR #103, opened 2020-06-30
Update caddy example config to address directive errors

eddiewebb:
I think the rewrite, cgi and gzip directives must be old, they have changed and the current example throws errors on Caddy 2.1

caddy.conf:3 - Error during parsing: Wrong argument count or unexpected line ending after 'rewrite'
caddy.conf:4: unrecognized directive: fastcgi
caddy.conf:10: unrecognized directive: gzip

This just updates to use the new directives.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -68,16 +68,14 @@ Caddy requires a very simple configuration in order for Flarum to work properly.
 ```
 www.example.com {
     root /var/www/flarum/public
-    rewrite {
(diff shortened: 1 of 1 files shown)


tankerkiller125:
The original caddy docs where written for Caddy v1 Caddy v2 is a very recent development and has only just recently been officially stable.

KyrneDev:
Tested this and it works!

PR #104, opened 2020-07-02
Updated wrong documentation about frontend routing.

x7airworker:
In the Discord I noticed that the frontend routing documentation is outdated.
The old JS Extender seems to be non functional.

### DIFF
--- a/docs/extend/routes.md
+++ b/docs/extend/routes.md
@@ -144,7 +144,7 @@ To register the route on the frontend, there is a `Routes` extender which works 
 ```jsx
 export const extend = [
   new Extend.Routes()
-    .add('/users', 'acme.users', <UsersPage />)
+    .add('acme.users', '/users', <UsersPage />)
 ];
 ```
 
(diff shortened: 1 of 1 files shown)


datitisev:
The bigger issue here is that the extender doesn't work. We should probably make an issue about that.

askvortsov1:
I 100% agree! But frankly I'm not sure when we're going to get around to implementing frontend extenders, and in the meantime, it would be good to have working docs

franzliedke:
If I look [at the implementation of this extender](https://github.com/flarum/core/blob/36eb5cc5fb4ed589b4f4f19171e07b86eb4cecef/js/src/common/extend/Routes.js#L4-L8), isn't this simply the wrong parameter order?

datitisev:
🤔 Apart from merging this PR to correct current documentation, I feel like we should switch the parameters around in core to reflect the Routes PHP extender.

x7airworker:
Yeah I also was a bit confused about the different order.

franzliedke:
@datitisev Fine for me, good suggestion. :+1:

askvortsov1:
This has been implemented elsewhere, thank you for the PR though!

PR #106, opened 2020-07-25
Provide Chinese translation of the document

scvoet:
As the title says, I am translating the document into Chinese. But at present I have only completed the language switch. In the future, I will gradually translate the documents into Chinese.


datitisev:
Hm, I think we'll probably want to move the locale stuff to their own files so that config.js doesn't become cluttered.

We can have sometihng like `.vuepress/locales/en.js` and just use `require('./locales/en.js')` 
```js
{
  // ...
  locales: {
    '/': require('./locales/en.js'),
    '/zh/': require('./locales/zh.js'),
  }
  // ...
}
```

Thoughts, core/docs people?

askvortsov1:
I would agree with that approach. My only concern is that no one on the core team is fluent in Chinese and is capable of maintaining / approving PRs. I'm definitely in favor of scaling to have other languages, especially considering how prominent international use of flarum is, but we should consider how we will ensure consistency across languages.

scvoet:
> 
> 
> Hm, I think we'll probably want to move the locale stuff to their own files so that config.js doesn't become cluttered.
> 
> We can have sometihng like `.vuepress/locales/en.js` and just use `require('./locales/en.js')`
> 
> ```js
> {
>   // ...
>   locales: {
>     '/': require('./locales/en.js'),
>     '/zh/': require('./locales/zh.js'),
>   }
>   // ...
> }
> ```
> 
> Thoughts, core/docs people?

I'd splitted it.

> 
> 
> I would agree with that approach. My only concern is that no one on the core team is fluent in Chinese and is capable of maintaining / approving PRs. I'm definitely in favor of scaling to have other languages, especially considering how prominent international use of flarum is, but we should consider how we will ensure consistency across languages.

I have no good way to solve the cross-language differences caused in the translation process. Maybe we can improve this problem to a certain extent by proofreading by Chinese users.

datitisev:
Thank you for your work, but some things (theme related, e.g. sidebar & nav) weren't working with your PR, and we have another one that contains actual Chinese documentation now (#107), so this PR is now obsolete.

Issue #87, opened 2020-03-07
Multi-language support?


Littlegolden:
I can translate documents into Chinese. Is there any plan to enable multi-language vuepress documents?

OrdinaryJellyfish:
VuePress has built-in support for internationalization: https://vuepress.vuejs.org/guide/i18n.html

datitisev:
I think we agreed internally on enabling internationalization when we are sure the translated documentation will be kept up-to-date. I cannot offer more details on this topic because I do not remember them 🙂.

luceos:
@datitisev if you can enable it and offer some simple instructions what to add where we can allow @Littlegolden to start translating. Do not translate docs about the frontend extending yet though..

datitisev:
What we'd have to do is add a `locales` key to `.vuepress/config.js`.

This is their example on their docs:

```js
module.exports = {
  locales: {
    // The key is the path for the locale to be nested under.
    // As a special case, the default locale can use '/' as its path.
    '/': {
      lang: 'en-US', // this will be set as the lang attribute on <html>
      title: 'VuePress',
      description: 'Vue-powered Static Site Generator'
    },
    '/zh/': {
      lang: 'zh-CN',
      title: 'VuePress',
      description: 'Vue 驱动的静态网站生成器'
    }
  }
}
```

Then we could create the documentation in a `docs/zh` folder (if we followed this example).

Default texts in the theme could then be changed with `themeConfig.locales`

```js
module.exports = {
  locales: { /* ... */ },
  themeConfig: {
    locales: {
      '/': {
        // text for the language dropdown
        selectText: 'Languages',
        // label for this locale in the language dropdown
        label: 'English',
        // Aria Label for locale in the dropdown
        ariaLabel: 'Languages',
        // text for the edit-on-github link
        editLinkText: 'Edit this page on GitHub',
        nav: [
          { text: 'Nested', link: '/nested/' , ariaLabel: 'Nested' }
        ],
        sidebar: {
          '/': [/* ... */],
          '/nested/': [/* ... */]
        }
      },
      '/zh/': {
        selectText: '选择语言',
        label: '简体中文',
        editLinkText: '在 GitHub 上编辑此页',
        nav: [
          { text: '嵌套', link: '/zh/nested/' }
        ],
        sidebar: {
          '/zh/': [/* ... */],
          '/zh/nested/': [/* ... */]
        }
      }
    }
  }
}
```

Littlegolden:
[...]

PR #107, opened 2020-07-26
Support Chinese Documentation

Littlegolden:
I haven't translated the /extend document yet as requested [before](https://github.com/flarum/docs/issues/87#issuecomment-619345895).

I found [the new PR]( #106 ) today and just realized the translation I almost forgot about.

As you guys discussed in that PR, I have categorized the configuration files, and I can expect that many different language translations will be PRed in the future, so I think the categorization will be very helpful to the developers involved.

Also, to prevent private information from being leaked, I unified the way the secret id used by the plugin is loaded, in such a way that the secret information (/config/secret.js) will always be stored in the local machine of the developer who eventually generates the static web page. 

BTW, I added a new plugin (@vuepress/plugin-container) to display the Flarum default directory structure (custom style) in  `/zh/install.html#url-rewrite`. The effect is like this (There's actually no change to the en files):
image

Oh, and the Contributor License Agreement, I have referred to the English version of various website agreement terms and the corresponding Chinese version (Adobe/Google/ etc.), and I can confirm that there is no error.

Above, I hope you like it！


datitisev:
❓ Perhaps we want to add a warning to the Chinese documentation that says it may be outdated?

--- 

This looks fine to me - apart from some things that I have fixed locally, including:

- revert secret.js (API key was provided to us, it's used in FOSS projects)
- fix issues with sidebar (`/extend/` must go before `/` in `en` sidebar)
- reorganize folder structure
    image

These things I have already fixed, see https://github.com/flarum/docs/commit/87c9ecae95c7b84847989f82308902efa9549b96.

If the rest of this PR looks fine, I'll close it and merge our `locales/zh` branch into master. The Chinese extending docs can be PR'd at another time.

Littlegolden:
> Perhaps we want to add a warning to the Chinese documentation that says it may be outdated?
Yeah, we need this. I have added an obsolete reminder to the homepage. https://github.com/flarum/docs/pull/107/commits/93ce635c9d9ecf5d373879249cc315f3f46d0595

> If the rest of this PR looks fine, we can merge it and put in the changes above afterwards. The Chinese extending docs can be PR'd at another time.

Okey, nice.

Littlegolden:
All done~

luceos:
Thank you @Littlegolden 👍

PR #110, opened 2020-07-31
Added an alternative to install through composer

Sarveshmrao:
Added an alternative to install through composer: just run the below composer line in your computer and install the generated vendor for to your server through ftp.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -22,7 +22,7 @@ Before you install Flarum, it's important to check that your server meets the re
 ::: tip Shared Hosting
 At this stage, it's not possible to install Flarum by downloading a ZIP file and uploading the files to your web server. This is because Flarum uses a dependency-management system called [Composer](https://getcomposer.org) which needs to run on the command line.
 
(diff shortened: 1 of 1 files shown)


franzliedke:
@flarum/docs Do we want to support this?

I think this was discussed before - and IIRC, we did not want to, as there are no guarantees that the prerequisites checked to be present when executing `composer` will be present on the production system as well.

askvortsov1:
I'm all fine with mentioning it as an option, but there should definitely be some very strongly worded disclaimers that it isn't an official installation method and that if the environments don't match, unexpected issues may occur.

luceos:
I'm becoming increasingly less a fan of openly promoting alternate solutions, including pockethold. Can't we create this under a FAQ or separate "trouble installing" page?

askvortsov1:
Agreed. IMO composer 2 took care of the last remaining reasons to use alternatives, and pocket hold isn't fully maintained. If we eventually add some kind of UI for composer we can revisit this, but until then I think we're fine.

askvortsov1:
Superceded by #194

PR #113, opened 2020-08-09
fix caddy root directive

jxsl13:
and update to newer php version


franzliedke:
For reference: https://caddyserver.com/docs/caddyfile/directives/root

PR #118, opened 2020-08-23
Add documentation for console (both for users and developers)

askvortsov1:


docs/console.md
```diff
@@ -0,0 +1,57 @@
...
+
+`php flarum migrate`
+
+Runs all outstanding migrations. This should be used when an extension that modifies the database is added or updated.
+
+### migrate:reset
+
+`php flarum migrate:reset --extension [extension_id]`
+
+Reset all migrations for an extension. This is mostly used by extension developers, but on occasion, you might need to run this if you are removing an extension, and want to clear all of its data from the database.
```
datitisev, line 51: You need the extension to be installed if you want to reset its migrations. I would add this detail here.

docs/console.md
```diff
@@ -0,0 +1,57 @@
...
+
+### generate:migration
+
+`php flarum generate:migration --extension [extension_id] [name]`
+
+Generates an empty migration file with a given name for a given extension. This is mostly useful for extension developers.
```
datitisev, line 57: I think you need to pass the full path to this command? If I remember correctly, it still tries to use the extensions/ directory.

docs/console.md
```diff
@@ -0,0 +1,57 @@
...
+
+`php flarum migrate:reset --extension [extension_id]`
+
+Reset all migrations for an extension. This is mostly used by extension developers, but on occasion, you might need to run this if you are removing an extension, and want to clear all of its data from the database. Please note that the extension in question must currently be installed (but not necessarily enabled) for this to work.
+
+### generate:migration
+
+`php flarum generate:migration --extension [composer_extension_name] [name]`
+
+Generates an empty migration file with a given name for a given extension. This is mostly useful for extension developers. Please note that you should provide the composer extension name instead of the composer id (e.g. `flarum/flags` instead of `flarum-flags`).
```
luceos, line 57: @flarum/core this command is lacking I think. Currently the generate:migration command will save the database file to the vendor extension migrations directory. Not a single extension developer I know of uses the vendor directory for development. For now we should not document this command, maybe we need to drop the command as well.

luceos (changes requested):

docs/console.md
askvortsov1: Well, if it's symlinked, shouldn't it get created in the right place?

luceos: Who says the extension is already installed? But yeah, okay, if it is using symlinks should work.

If you ask me, I'd rather create a developer extension/package (similar to what @datitisev created) which we can ship for developers as a companion toolkit. This kit would not target the vendor directory but the configured workbench directory.

So, question remains. Do we want to keep this docs entry, this is the only developer relating command we have.

askvortsov1: I'd be fine removing it tbh. I'd also like to add in a little blurb about scheduling commands via fof console.

tankerkiller125: fof console isn't a core feature though so I don't think it makes sense to include it.

askvortsov1: Fair, although we do document some unofficial utils occasionally. Maybe just a quick info thing? (not a full docs section). And I did open an issue for discussion of maybe making that a core feature: https://github.com/flarum/core/issues/2281

luceos (approved):

PR #121, opened 2020-08-28
Get started with Beta.14 upgrade guide

franzliedke:
Let's get this ready so that extension developers can prepare themselves for Beta.14.

Since this will be a relatively complex upgrade for those not familiar with the internals of Mithril and its different versions, let's go above and beyond in helping them understand what needs to be done.

Refs flarum/core#2201.

### Open TODOs:

- [ ] Finish section on backend changes
- [ ] Mention removal of auth-* extensions, see flarum/core#2006

### DIFF
--- a/docs/extend/update-b14.md
+++ b/docs/extend/update-b14.md
@@ -702,16 +702,15 @@ Considering you have to do the changes anyway, why not do them now?
 
 ### Removals
 
-Warnings to users
-
-No more laravel helpers
-Symfony translator interface should be used, not Laravel's
-Do NOT use the old callback notation for configuring view namespaces. This will break all extensions that boot after your extension. The view extender must be used.
-
-*TODO*
-
-- The following events [deprecated in Beta 13](https://github.com/flarum/core/commit/4efdd2a4f2458c8703aae654f95c6958e3f7b60b) have been removed:
+- Do NOT use the old closure notation for configuring view namespaces. This will break all extensions that boot after your extension. The `View` extender MUST be used instead.
+- app()->url() will no longer work: [`Flarum\Http\UrlGenerator`](routes.md) should be injected and used instead. An instance of `Flarum\Http\UrlGenerator` is available in `blade.php` templates via `$url`.
+- As a part of the Laravel 6 upgrade, the [`array_` and `str_` helpers](https://laravel.com/docs/6.x/upgrade#helpers) have been removed.
+- The Laravel translator interface has been removed; the Symfony translator interface should be used instead: `Symfony\Component\Translation\TranslatorInterface`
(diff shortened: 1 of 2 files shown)


docs/extend/update-b14.md
```diff
@@ -0,0 +1,98 @@
...
+When you have taken care of the changes above, you should be good to go.
+For the following changes, we have bought you time until the stable release.
+Considering you have to do the changes anyway, why not do them now?
+
+*TODO*
+
+- `this.props` -> `this.attrs`
+- static `initProps()` -> static `initAttrs()`
+- `m.prop` -> `m.stream`
+- `m.withAttr` -> `withAttr` with import
```
franzliedke, line 75: This is my proposal for a minimal BC layer. These... 

a) are trivial to add and maintain,
b) will be removed for the stable release, and
c) are repetitive and frequent, so avoiding them now should help the extension ecosystem.

docs/extend/update-b14.md
```diff
@@ -0,0 +1,98 @@
...
+  - register routes with page class, not instance
+    - special case when passing props
+  - `<a href={url} config={m.route}>` -> `<a route={url}>`
+- AJAX
+  - `m.request({...})` -> `data:` key split up into `body:` and `params:`
+  - `m.deferred` -> native `Promise`
+- Redrawing
+  - `m.redraw(true)` -> `m.redraw.sync()`
+  - `m.redraw.strategy('none')` -> `e.redraw = false` in event handler
+  - `m.lazyRedraw()` -> `m.redraw()`
```
franzliedke, line 59: These will need to be explained thoroughly. But for now I wanted to collect the steps here. Is there anything missing (here or in the `Optional` section below) related to the Mithril 2.0 upgrade?

datitisev:
I think we should mention that having empty strings as Element attributes will now error - probably not many people do this but I had some code and it took me a while until I was able to locate the problem.

Some code like this (looks ugly but I'd rather have it inline LOL)

image

`''` can be replaced with `null` and it'll work fine.

docs/extend/update-b14.md
datitisev: Not sure what you mean with this comment? Do you want to remove all of these?

docs/extend/update-b14.md
datitisev: I'd add that you can no longer use a `children` attribute unless the component extracts it before passing it as like `{...attrs}` or something.

docs/extend/update-b14.md
```diff
@@ -0,0 +1,388 @@
...
+
+```js
+class ChildComponent extends Component {
+  view() {
+    return <p>{this.attrs.counter}</p>;
+  }
+}
+
+class ParentComponent extends Component {
+  init() {
```
datitisev, line 173: This should be changed to `oninit` and call `super.oninit(...)`

KyrneDev (commented):
Looks good in current state. Only issue I see is there are a few comments before JS code blocks that are annotated as mithril 0.2 when it's actually 2.0

askvortsov1 (approved):

Issue #122, opened 2020-09-04
Introduce official code-generated documentation for PHP and JS


askvortsov1:
It would be nice to convert http://api.flarum.dev/ into official documentation that we could reference from the vuepress docs.

datitisev:
Reference in what whay? How would the official one differ from my unofficial one?

askvortsov1:
It wouldn't differ, but we would want to keep it updated. And when I say reference, I mean we could link to docblocks from other parts of the documentation.

datitisev:
>  I mean we could link to docblocks from other parts of the documentation.

Wouldn't this require some integration with vuepress? Or would it just be a link that goes to the API doc page?

franzliedke:
Links would be fine, and that's what I was always hoping to have for the extenders (i.e. describe them shortly in the prose docs, and then link off to the API docs that have all the details thanks to docblocks).

@datitisev Oh, or were you worried about having versioned URLs and keeping them updated in the docs?

PR #128, opened 2020-09-18
Links

spekulatius:
Hey @luceos 

noticed a two more broken links: One in the readme here on the repo and one on the site.

Cheers!

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -12,6 +12,6 @@ Flarum is the combined successor of [esoTalk](https://github.com/esotalk/esoTalk
 
 * **Beautiful and responsive.** This is forum software for humans. Flarum is carefully designed to be consistent and intuitive across platforms, out-of-the-box.
 
-* **Powerful and extensible.** Customize, extend, and integrate Flarum to suit your community. Flarum’s architecture is amazingly flexible, with a [powerful Extension API](https://docs.flarum.org/extend/).
+* **Powerful and extensible.** Customize, extend, and integrate Flarum to suit your community. Flarum’s architecture is amazingly flexible, with a [powerful Extension API](/extend/).
 
 * **Free and open.** Flarum is released under the [MIT license](https://github.com/flarum/flarum/blob/master/LICENSE).
--- a/docs/README.md
+++ b/docs/README.md
@@ -12,6 +12,6 @@ Flarum is the combined successor of [esoTalk](https://github.com/esotalk/esoTalk
 
 * **Beautiful and responsive.** This is forum software for humans. Flarum is carefully designed to be consistent and intuitive across platforms, out-of-the-box.
 
-* **Powerful and extensible.** Customize, extend, and integrate Flarum to suit your community. Flarum’s architecture is amazingly flexible, with a [powerful Extension API](https://github.com/flarum/docs/tree/master/docs/extend).
+* **Powerful and extensible.** Customize, extend, and integrate Flarum to suit your community. Flarum’s architecture is amazingly flexible, with a [powerful Extension API](https://docs.flarum.org/extend/).
 
 * **Free and open.** Flarum is released under the [MIT license](https://github.com/flarum/flarum/blob/master/LICENSE).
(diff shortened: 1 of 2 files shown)


docs/README.md
```diff
@@ -12,6 +12,6 @@ Flarum is the combined successor of [esoTalk](https://github.com/esotalk/esoTalk
 
 * **Beautiful and responsive.** This is forum software for humans. Flarum is carefully designed to be consistent and intuitive across platforms, out-of-the-box.
 
-* **Powerful and extensible.** Customize, extend, and integrate Flarum to suit your community. Flarum’s architecture is amazingly flexible, with a [powerful Extension API](/extend/).
+* **Powerful and extensible.** Customize, extend, and integrate Flarum to suit your community. Flarum’s architecture is amazingly flexible, with a [powerful Extension API](https://github.com/flarum/docs/tree/master/docs/extend).
```
askvortsov1, line 15: What's the reasoning behind this change? I think it would be better for this to be an internal link to the extension docs rather than the code. Once we introduce official API docs (#122), we could link there as well

askvortsov1 (changes requested):

docs/README.md
spekulatius: Oh, I see. I clicked on the link here in the repo and thought it's broken (not thinking about the relative host on the site). How about we make it an absolute link to the docs? Then it works for both.

spekulatius: I've tweaked it quickly to ensure it works on both @askvortsov1

askvortsov1 (approved):

docs/README.md
```diff
@@ -12,6 +12,6 @@ Flarum is the combined successor of [esoTalk](https://github.com/esotalk/esoTalk
 
 * **Beautiful and responsive.** This is forum software for humans. Flarum is carefully designed to be consistent and intuitive across platforms, out-of-the-box.
 
-* **Powerful and extensible.** Customize, extend, and integrate Flarum to suit your community. Flarum’s architecture is amazingly flexible, with a [powerful Extension API](/extend/).
+* **Powerful and extensible.** Customize, extend, and integrate Flarum to suit your community. Flarum’s architecture is amazingly flexible, with a [powerful Extension API](https://docs.flarum.org/extend/).
```
datitisev, line 15: I don't think this change should be here - the current link works fine and we shouldn't be hardcoding the URL.

datitisev (changes requested):

docs/README.md
askvortsov1: Sorry for the confusion (I'm jumping between a bunch of things today, and it's showing unfortunately :see_no_evil:). Wdym by "clicked on the link here in the repo"? Is there a use case for people acessing documentation links from within the github repo?

docs/README.md
Littlegolden: Yes, we shouldn't be hardcoding the URL. and it's better to keep using relative links.

spekulatius: Yeah, it works on the site. Just not on the repo https://github.com/flarum/docs. But I can undo it.

Littlegolden: Because after the document is modified, if someone need a local online preview of the site, the relative link is required to open the latest local page.

docs/README.md
spekulatius: Hey @askvortsov1, I know that feeling. I jump around too much as well.  by "clicked on the link here in the repo" here https://github.com/flarum/docs. It's the same page, in different context.

docs/README.md
spekulatius: Let's undo it then 👍️

docs/README.md
spekulatius: I've tweaked it back to allow local editing over fixing the link on GitHub (makes sense).

docs/README.md
spekulatius: I've undone the change. It was just a quick thing I've noticed and added to the PR along the way anyways.

datitisev (approved):
Now, with just the `/` being added to the extend link, this looks good to me.

spekulatius:
> Now, with just the `/` being added to the extend link, this looks good to me.

Awesome! Thanks for checking!

askvortsov1 (approved):

askvortsov1:
Thank you!

spekulatius:
Thank you all for reviewing it - much more complicated than I expected haha

PR #131, opened 2020-10-09
Document Authorization, Visibility Scoping, Models, and Permissions

askvortsov1:
This is some of the stuff I struggled with most when getting started, so would be nice to have it documented, even if we redo the entire policy system in beta 15!


clarkwinkelmann (approved):
Looks good to me.

My only worry is that this might be too advanced for many. I personally understand all of the content but I'm sure it will be difficult to grasp for newcomers.

But since we're going to re-do this anyway, any documentation is better than no documentation for now.

PR #132, opened 2020-10-09
Add validation documentation

### LINKED ISSUE #169: Document validation

askvortsov1:


datitisev:
Without actually reading this, 3 out of the 4 code blocks have a 2-space indentation. This should be changed to 4 spaces for consistency & better readability.

docs/extend/validation.md
```diff
@@ -0,0 +1,179 @@
...
+use Flarum\Extend;
+use Flarum\Foundation\Event\Validating;
+use Flarum\User\UserValidator;
+use Illuminate\Support\Str;
+
+return [
+    // Register extenders here
+    (new Extend\Event)->listen(Validating::class, function(Dispatcher $events) {
+        $events->listen(Validating::class, function(Validating $event) {
+        // This modification should only apply to UserValidator
```
clarkwinkelmann, line 149: There's an issue with indentation here. This should be one level deeper.

There's also a `return` at line 156 that needs one additional level.

And the whole 162-170 also needs one more level.

docs/extend/validation.md
```diff
@@ -0,0 +1,179 @@
...
+    }
+}
+```
+
+If it fails validation, a `Illuminate\Validation\ValidationException` will be thrown with the validation errors, so that Flarum's error handling system can optimally present them to the user.
+
+Remember that you can turn an Eloquent instance into an associative array of attributes via:
+
+- The `$instance->getAttributes()` method if the instance hasn't been saved.
+- The `$instance->getDirty()` method if the instance has been saved.
```
clarkwinkelmann, line 35: I'm not sure if that's correct or if I correctly understand what you wanted to describe.

- `getAttributes()` will get all attribute values as they are currently present on the model, saved or not
- `getDirty()` will get all attribute values that have been modified and not yet saved. It works for new or existing models
- `getChanges()` will get all attributes that were modified and are now saved
- `getOriginal()` will get all attribute values that were retrieved from the database

I think it would also be important here to mention that we commonly validate data before pushing it into the model. I think most extensions do that, and directly validate the data from the request.

clarkwinkelmann (commented):
Globally ok. Few notes.

Also, I recognize that validator extender example :stuck_out_tongue:

docs/extend/validation.md
askvortsov1: How did I miss that...

docs/extend/validation.md
askvortsov1: To be honest, this is more an instance of me not understanding part of Eloquent. Learned something new today!

askvortsov1:
> Also, I recognize that validator extender example

First thing I did when trying to revamp documentation was look through your recent Discuss posts :bow:

Oh, and that Eloquent stuff you corrected me on just now is going right in there too :grinning:

docs/extend/validation.md
```diff
@@ -31,8 +31,12 @@ If it fails validation, a `Illuminate\Validation\ValidationException` will be th
...
 Remember that you can turn an Eloquent instance into an associative array of attributes via:
 
-- The `$instance->getAttributes()` method if the instance hasn't been saved.
-- The `$instance->getDirty()` method if the instance has been saved.
+- `$instance->getAttributes()` will get all attribute values as they are currently present on the model, saved or not
+- `$instance->getDirty()` will get all attribute values that have been modified and not yet saved. It works for new or existing models
+- `$instance->getChanges()` will get all attributes that were modified and are now saved
+- `$instance->getOriginal()` will get all attribute values that were retrieved from the database
+
+Also, keep in mind that it's generally preferable to validate data before pushing it into the model instance.
```
clarkwinkelmann, line 39: I think it would be nice to include a small example here. Maybe:

```php
<?php

use Illuminate\Support\Arr;
use Psr\Http\Message\ResponseInterface;
use Psr\Http\Message\ServerRequestInterface;
use Psr\Http\Server\RequestHandlerInterface;

class MyController implements RequestHandlerInterface
{
    protected $validator;
    public function __construct(UserValidator $validator) {
        $this->validator = $validator;
    }

    public function handle(ServerRequestInterface $request): ResponseInterface
    {
        $attributes = Arr::get($request->getParsedBody(), 'data.attributes', []);

        $this->validator->assertValid($attributes);

        // Do something with the attributes
        // Access them with for example Arr::get($attributes, 'email')
    }
}
```

Also worth noting somehow:

When validating an array, only keys that are present in the validator will be checked. Other will be left as-it. You should only read values from the array that were validated. To get an array of values that ran through the validator, you can use `Arr::only($attributes, array_keys($validator->getRules()))`.

A validator based on Flarum's `AbstractValidator` will only validate keys which are present in the `$data` parameter. This means `required` rules won't be checked if the data is absent. This is designed to handle API requests where each attribute will be handled separately and ignored when absent. To validate all data including missing data, you need to use a Laravel validator.

imorland:
Closing as stale. This PR (2020) adds files under `docs/.vuepress/`, which no longer exists — the docs site migrated from VuePress to Docusaurus. Validation documentation can be revisited fresh against the current 2.x structure if needed.

PR #134, opened 2020-10-15
Upgrade VuePress to v1.7.1

lhsazevedo:
Upgrade VuePress from 1.5.2 to ^1.7.1 (greater or equal than 1.7.1 and less than 2.0.0).


datitisev:
I think updating makes sense now - I think a version after 1.5.2 had bugs or something so we didn't update. I'll test locally before merging.

datitisev (approved):

PR #135, opened 2020-10-15
Update accent color

lhsazevedo:
Use primary color from [Flarum Community](https://discuss.flarum.org) as accent color.

**Before:**
before

**After:**
after


datitisev:
I think this new accent color makes the contrast between the background and the text worse... I don't think it needs to be changed.

PR #142, opened 2020-10-24
Update Chinese Translations

Littlegolden:
1. Sync with 2 commits behind: 8ba07e5, d850dc9
2. Correct url slug within `contributing.md`, `install.md`, `troubleshoot.md`.
3. Update outdate-tip component.
4. Add `/extend` README.


tankerkiller125 (approved):
I don't see any glaring issues as far as any of the english, links, etc. goes.

datitisev:
I don't see why we need those Vue components that are just in Chinese?

Littlegolden:
I have not figured out a better way to reuse the "translation may not be updated in time" prompt for now, so I used component.  Because only Chinese translation is served at present. Maybe use js to output prompt sentences in different languages is a better way if there is another language translation?

askvortsov1:
Makes sense to me, thanks for putting this together!

> I have not figured out a better way to reuse the "translation may not be updated in time" prompt for now, so I used component. Because only Chinese translation is served at present. Maybe use js to output prompt sentences in different languages is a better way if there is another language translation?

That's a good idea, but I feel like it's better to worry about that when we have docs in other languages (and someone as awesome as you to maintain them :hugs: :smile: (seriously, thank you so much for doing this!)).

docs/.vuepress/components/processing.vue
```diff
@@ -0,0 +1,7 @@
+<template>
+    <div class="notification-container">
+        <div class="processing-tip">
+            <p>⚡当前语言可能更新不及时，最新信息请以<a href="/" target="_blank" rel="noopener noreferrer">英文文档</a>为准。如果翻译有问题，请<a href="https://github.com/flarum/docs/issues" target="_blank" rel="noopener noreferrer">告诉我们</a>。</p>
+        </div>
+    </div>
+</template>
```
askvortsov1, line 7: Might be good to ensure an ending newline for sake of convention, but that's super minor and nitpicky :grin:

askvortsov1 (approved):

docs/.vuepress/components/processing.vue
Littlegolden: Nothing to it~ 😄

askvortsov1 (approved):

PR #144, opened 2020-10-30
Proposal: Navigation and base font update

lhsazevedo:
Attemps to match Flarum's brand identity by updating the styles of nav links, search box, base font (Mukta) and base body color.

Before/After:
compariso


askvortsov1:
I like this! It would be nice to still have links bolded though

luceos (approved):
@lhsazevedo that's so awesome, thanks!!

@askvortsov1 discuss doesn't have bolded links either, I'm totally okay with this!

PR #147, opened 2020-11-03
Updated instructions to fix permissions errors

nina-py:
Add-on to the PR that updates the error message during installation - flarum/core#2435. 

Fixes flarum/core#2327

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -94,7 +94,9 @@ chmod 775 -R /path/to/directory
 
 If after completing these steps, Flarum continues to request that you change the permissions you may need to check that your files are owned by the correct group and user. 
 
-By default, in most Linux distributions `www-data` is the group and user that both PHP and the web server operate under. You can change the folder ownership in most linux operating systems by running `chown -R www-data:www-data foldername/`. 
+By default, in most Linux distributions `www-data` is the group and user that both PHP and the web server operate under. You can change the folder ownership in most Linux operating systems by running `chown -R www-data:www-data foldername/`. 
+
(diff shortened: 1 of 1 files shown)


askvortsov1 (changes requested):
Users that need this documentation are likely to be unfamiliar with server administration, so I think it might be better to add a bit more explanation of what folder ownership and these commands mean (or add some links to sources that explain it well)

nina-py:
@askvortsov1 , sure, I'll give it a go! Do you think it would be a good idea to mention that for some Linux distros you need to prefix these commands with `sudo` as well?

askvortsov1:
> @askvortsov1 , sure, I'll give it a go! Do you think it would be a good idea to mention that for some Linux distros you need to prefix these commands with `sudo` as well?

Perhaps. A bit about Windows might be nice too. I do think that links for background info might be the way to go: that way we can keep our docs slim and focused on the actual flarum stuff (as we do in other parts of the docs).

nina-py:
Agree re slim answers with links - there appears to be a canonical answer on SuperUser for Windows users that we could use: https://superuser.com/questions/106181/equivalent-of-chmod-to-change-file-permissions-in-windows. For Linux file permissions in general, I could link to this tutorial from the Linux Foundation: https://www.linux.com/training-tutorials/understanding-linux-file-permissions/.

askvortsov1:
I also like https://www.thegeekdiary.com/understanding-basic-file-permissions-and-ownership-in-linux/. And the Linux foundation needs to work on their css :rofl:

askvortsov1 (approved):

tankerkiller125 (approved):

PR #153, opened 2020-12-18
AdminUX Docs

KyrneDev:
Adds info on how to use the new `ExtensionData` API and the composer.json metadata.

### DIFF
--- a/docs/extend/admin.md
+++ b/docs/extend/admin.md
@@ -91,10 +91,10 @@ app.initializers.add('interstellar', function(app) {
     
         return (
           <div className="Form-group">
-            <h1>You are Rocketman!</h1>
+            <h1> {app.translator.trans('acme-interstellar.admin.you_are_rocket_man_label')} </h1>
             <label className="checkbox">
               <input type="checkbox" bidi={this.setting('acme-interstellar.rocket_man_setting')}/>
-                {app.translator.trans('acme-interstellar.admin.rocket_man_label')}
+                {app.translator.trans('acme-interstellar.admin.rocket_man_setting_label')}
             </label>
           </div>
         );
--- a/docs/extend/admin.md
+++ b/docs/extend/admin.md
@@ -1,1 +1,159 @@
 # Admin Dashboard
+
+Beta 15 introduced a completely redesigned admin panel and frontend API. It is now easier than ever to add settings or permissions to your extension.
+
+Before beta 15, extension settings were either added in a `SettingsModal` or they added a new page for more complex settings. Now, every extension has a page containing info, settings, and the extension's own permissions.
+
+You can simply register settings, extend the base [`ExtensionPage`](https://api.docs.flarum.org/js/master/class/src/admin/components/extensionpage.js~extensionpage), or provide your own completely custom page.
+
+::: warning SettingsModal
+Settings added via the `SettingsModal` will contiune to work in beta 15, but this method **is now deprecated** and will be remove in the next release.
+:::
+
+## Extension Data API
+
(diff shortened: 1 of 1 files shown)


docs/extend/admin.md
```diff
@@ -1 +1,159 @@
...
+  options: {
+    'LOH': 'Liquid Fuel', // The key in this array is what the setting will be stored as in the database, the value is the label the admin will see (remember to use translations if they make sense in your context).
+    'RDX': 'Solid Fuel',
+  },
+  default: 'LOH',
+}
+```
+
+
+If you want to add something to the settings like some extra text or a more complicated input, you can also pass a callback as the first argument that returns JSX. This callback will be executed in the context of [`ExtensionPage`](https://api.docs.flarum.org/js/master/class/src/admin/components/extensionpage.js~extensionpage) and setting values will not be automatically serialized.
```
askvortsov1, line 81: Could you include an example, preferably one that binds to the setting val? Just a simple input would be fine

docs/extend/admin.md
```diff
@@ -1 +1,159 @@
...
+      }, 
+      'start', // Category permission will be added to on the grid
+      95 // Optional: Priority
+    );
+});
+```
+
+### Chaining Reminder
+
+Remember these functions call all be chained like:
```
askvortsov1, line 115: call => can

askvortsov1: Also can you put ellipses inside of all the `register()` like `registerSetting(...)`

docs/extend/admin.md
```diff
@@ -1 +1,159 @@
...
+
+  app.extensionData
+    .for('acme-interstellar')
+    .registerPage(StarPage);
+});
+```
+
+This page will be shown instead of the default.
+
+You can extend the [`ExtensionPage`](https://api.docs.flarum.org/js/master/class/src/admin/components/extensionpage.js~extensionpage) or extend the base `Page` and design your own!
```
askvortsov1, line 146: Maybe show a basic example of extending ExtensionPage to replace the settings content?

docs/extend/admin.md
```diff
@@ -1 +1,159 @@
...
+
+In beta 15, extension pages make room for extra info which is pulled from extensions' composer.json.
+
+| Description                       | Where in composer.json                 |
+| --------------------------------- | -------------------------------------- |
+| discuss.flarum.org discussion link | "forum" key inside "support"           |
+| Documentation                     | "docs" key inside "support"            |
+| Support (email)                   | "email" key inside "support"           |
+| Website                           | "homepage" key                         |
+| Donate                            | "funding" key block (see [flarum/tags](https://github.com/flarum/tags/blob/master/composer.json#L13-L18)) |
```
askvortsov1, line 158: Note that only the link in the first funding block is selected

docs/extend/admin.md
```diff
@@ -1 +1,159 @@
...
+});
+```
+
+This page will be shown instead of the default.
+
+You can extend the [`ExtensionPage`](https://api.docs.flarum.org/js/master/class/src/admin/components/extensionpage.js~extensionpage) or extend the base `Page` and design your own!
+
+## Composer.json Metadata
+
+In beta 15, extension pages make room for extra info which is pulled from extensions' composer.json.
```
askvortsov1, line 150: Add a link to Composer's docs for `composer.json`

askvortsov1 (changes requested):

askvortsov1 (changes requested):
The only concern I have at this point is the hard-coded text. Not sure if that shows up anywhere else in docs but translations are always preferable IMO.

PR #155, opened 2020-12-22
Update example in docs for Caddy

Telesphoreo:
Minor fixes to make Flarum display properly on Caddy. Without these, you will see an unformatted page that says "Something went wrong while trying to load the full version of this site. Try hard-refreshing this page to fix the error." This fixes that error


tankerkiller125:
Duplicate of #140

PR #148, opened 2020-11-05
Update Chinese Docs & Fix Typo

Littlegolden:
1. Add Chinese docs(extend/start.md)
2. Correct the reversed component name.
3. Add Chinese font.
4. Delete wrong sentences in `zh/contributing.md`. (found by @smyhw)

PR #160, opened 2021-01-23
Update install.md

lvzx123:
在php版本要求中增加“不支持8.x”
将/您的/论坛/路径/根据汉语习惯修改成/[您的论坛路径]/


Littlegolden:
1. I agree with adding the version prompt, we can remove it after flarum supports php8.

2. `[]` indicates optional, but in general, this path is indispensable.

askvortsov1:
@Littlegolden @lvzx123 Do we still need this with https://github.com/flarum/docs/pull/148?

lvzx123:
> @Littlegolden @lvzx123 Do we still need this with #148?

#148 is better

PR #163, opened 2021-03-02
add reference to git as the indirect dependency

plluksie:
git is needed to avoid error:
  [RuntimeException]
  git was not found in your PATH, skipping source download


Littlegolden (approved):

luceos:
Git is required by composer. You can't use composer if you don't have git. If you however install Flarum without composer (by uploading after having used composer locally) you won't need git. So I think we'll just leave it as is. 

Thanks for the consideration.

PR #178, opened 2021-03-15
Update Chinese docs & some improvement

Littlegolden:
1. i18n for Outdated Tip
2. Sync English docs: dfa3d7b, 14c434e
3. Add default directory structure (docs/install.md)


docs/install.md
```diff
@@ -109,6 +109,27 @@ You should never set any folder or file to permission level `777`, as this permi
...
+│
+├── vendor _(**flarum core, extensions**)_
+├── .nginx.conf _(**Nginx pseudo-static file**)_
+├── composer.json
+├── extend.php
+├── flarum
+├── LICENSE
+├── README.md
+└── site.php
+:::
```
Littlegolden, line 131: Add default directory structure

docs/.vuepress/public/js/i18n.js
```diff
@@ -0,0 +1,82 @@
...
+            processing_tip = 'The translation of extend documents is still a work in progress, please refer to the <a>English document</a> for the latest information. If you have problems with the translation, please <a>let us know</a>.';
+            is_processing = false;
+            break;
+        case 'zh':
+            outdated_tip = '此页面的翻译可能更新不及时，最新信息请以<a>英文文档</a>为准。如果翻译有问题，请<a>告诉我们</a>。';
+            processing_tip = '当前文档仍在翻译中（进度 6/23），有关最新信息，请查阅<a>英文文档</a>。如果翻译有问题，请<a>告诉我们</a>。';
+            is_processing = true;
+            break;
+        case 'tr':
+            outdated_tip = 'The translation of this page may not be up to date, please refer to the <a>English document</a> for the latest information. If there is a problem with the translation, please <a>let us know</a>.';
```
askvortsov1, line 16: Would it be possible to use translated text for these?

docs/install.md
```diff
@@ -109,6 +109,27 @@ You should never set any folder or file to permission level `777`, as this permi
 
 By default Flarum's directory structure includes a `public` directory which contains only publicly-accessible files. This is a security best-practice, ensuring that all sensitive source code files are completely inaccessible from the web root.
 
+::: vue
```
askvortsov1, line 112: I'm not sure this is the best place for this. Maybe under 'Installing' would be better? Also, this might be better in a separate PR.

docs/zh/troubleshoot.md
```diff
@@ -1,7 +1,3 @@
-<template>
-  <outdated class="blue"></outdated>
-</template>
-
 # 故障排查
```
askvortsov1, line 1: Could these be removed from other language files so they don't appear twice?

askvortsov1 (changes requested):
Thank you for putting this together! Sorry for the delay in reviewing.

docs/zh/troubleshoot.md
Littlegolden: Other language files haven't used this template yet, so the notification will appear once.

docs/install.md
Littlegolden: sonds good, I'll do that

docs/.vuepress/public/js/i18n.js
Littlegolden: I don't speak Turkish, we can let @DursunCanPoyraz translate it later

docs/install.md
Littlegolden: this change has been reverted

docs/zh/troubleshoot.md
askvortsov1: Isn't it being used in the Italian version?

Littlegolden: No, it isn't. I searched this template code in Italian doc, and found nothing.

askvortsov1 (approved):

davwheat (approved):
I haven't tested it locally, but it all looks good to me.

PR #192, opened 2021-04-08
Reorganize models and data docs

askvortsov1:
- Split it into 2 separate pages, as there's a LOT of info crammed in one article right now
- Add a flowchart diagram to the API process description
- A few other small updates and improvements


docs/extend/api.md
```diff
@@ -0,0 +1,331 @@
...
+#### Searching and Filtering
+
+Read our [searching and filtering](search.md) guide for more information!
+
+### Showing a Resource
+
+For the controller that shows a single resource, extend the `Flarum\Api\Controller\AbstractShowController` class. Like for the list controller, you need to specify the `$serializer` you want to use to serialize your models, and implement a `data` method to return a single model.
+We'll learn about serializers [in just a bit](#serializers). 
+
+
```
davwheat, line 133: Is there meant to be a new paragraph between 130 and 131? We can also remove one of the line breaks on 132/133

docs/extend/models.md
```diff
@@ -0,0 +1,311 @@
...
+discussion.delete().then(done);
+```
+
+## Backend Models vs Frontend Models
+
+Often, backend and frontend models will have similar attributes and relationships. This is a good pattern to follow, but isn't always true.
+
+The attributes and relationships of backend models are based on the **database**. Each column in the model's table will map to an attribute on the backend model.
+
+The attributes and relationships of frontend models are based on the output of [**API Serializers**](api.md). These will be covered more in depth in the next article, but it's worth that a serializer could output all, any, or none of the backend model's attributes, and the names under which they're accessed might be different in the backend and frontend.
```
davwheat, line 309:
```suggestion
The attributes and relationships of frontend models are based on the output of [API Serializers](api.md). These will be covered more in depth in the next article, but it's worth that a serializer could output all, any, or none of the backend model's attributes, and the names under which they're accessed might be different in the backend and frontend.
```

davwheat (approved):

Issue #190, opened 2021-04-05
Mailgun driver requirements


clarkwinkelmann:
The Email page should list the requirements for Mailgun driver https://docs.flarum.org/mail.html

Similar to #83 

I think we should recommend running `composer require guzzlehttp/guzzle:^6.0|^7.0`. Running `composer require guzzlehttp/guzzle` might leave you blocked with this error when installing extensions https://discuss.flarum.org/d/24684-can-only-install-one-of-guzzlehttpguzzle

PR #193, opened 2021-04-08
Document mailgun's need for guzzle

askvortsov1:
Fixes https://github.com/flarum/docs/issues/190


clarkwinkelmann (approved):
Not sure about using `<code>` for the name of the package. Maybe just say Guzzle with uppercase?

Looks good otherwise

PR #200, opened 2021-04-20
Split model visibility, document `where` vs `orWhere`

askvortsov1:


docs/extend/model-visibility.md
```diff
@@ -0,0 +1,240 @@
+# Model Visibility
+
+This article concerns authorization, and uses some concepts from the [authorization](authorization.md) system. You should familiarize yourself with that first.
+
+When a user visits the **All Discussions** page, we want to quickly show them the recent discussions that the user has access to.
+We do this via the `whereVisibleTo` method, which is defined in `Flarum\Database\ScopeVisibilityTrait`, and available to [Eloquent models and queries](https://laravel.com/docs/6.x/queries) through [Eloquent scoping](https://laravel.com/docs/6.x/eloquent#local-scopes).
```
luceos, line 6: This links to the Laravel 6 docs, we're on 8.
```suggestion
We do this via the `whereVisibleTo` method, which is defined in `Flarum\Database\ScopeVisibilityTrait`, and available to [Eloquent models and queries](https://laravel.com/docs/8.x/queries) through [Eloquent scoping](https://laravel.com/docs/8.x/eloquent#local-scopes).
```

luceos (changes requested):
Now that I read all this, it's even more obvious how confusing this will be to developers. For 2.x we seriously need to improve the logic behind opening up private discussions; aka wrap each extender into their own `orWhere` or something.

Anyway good job getting this in print :P

SychO9 (approved):

PR #203, opened 2021-04-26
Spanish Docs

IamDarkle:
Translation of Flarum documents into Spanish


askvortsov1:
This is fantastic, thank you! Could you also add an entry for Spanish to https://github.com/flarum/docs/blob/master/docs/.vuepress/public/js/i18n.js?

IamDarkle:
Sure! I forgot that entry, it's ready now

askvortsov1 (approved):

PR #206, opened 2021-04-27
config.md: Add site maintenance mode flag

t4top:
It is possible to enable or disable site maintenance mode with `offline` key in [config.php](https://github.com/flarum/core/blob/c119731e652fcfb96e1bb410e0550c6fbb3dc445/src/Foundation/Config.php#L42) but this was not documented. This PR adds this missing key into the docs.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -12,7 +12,7 @@ Here's a quick overview of what everything means with an example file:
 ```php
 <?php return array (
   'debug' => false, // enables or disables debug mode, used to troubleshoot issues
-  'offline' => false, // enables or disables site maintenance mode
+  'offline' => false, // enables or disables site maintenance mode. This makes your site inaccessible to all users (including admins).
   'database' =>
   array (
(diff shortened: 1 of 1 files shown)


docs/config.md
```diff
@@ -12,6 +12,7 @@ Here's a quick overview of what everything means with an example file:
 ```php
 <?php return array (
   'debug' => false, // enables or disables debug mode, used to troubleshoot issues
+  'offline' => false, // enables or disables site maintenance mode
```
askvortsov1, line 15:
```suggestion
  'offline' => false, // enables or disables site maintenance mode. This makes your site inaccessible to all users (including admins).
```

Thanks, this is somewhat of an "unknown" feature. I think it would be good to clarify what this does, so people don't get confused (at least until https://github.com/flarum/core/issues/1732 gets fixed).

askvortsov1 (changes requested):

askvortsov1 (approved):

PR #325, opened 2021-05-12
Update for stable

### LINKED ISSUE #172: Document Filesystem Extensibility

### LINKED ISSUE #201: Document extension assets
Essentially, just https://discuss.flarum.org/d/26928-how-to-manage-positions-into-extensions/6 and an explanation of the publish assets command

askvortsov1:
See commit titles for details

Closes #201 
Closes #172
Closes https://github.com/flarum/core/issues/2202


docs/extend/i18n.md
```diff
@@ -172,28 +178,27 @@ Of course, you can give a parameter any name you like &mdash; you could use `<fr
...
 
 ```yaml
-choose_primary_placeholder: "Choose a primary tag|Choose {count} primary tags"
+choose_primary_placeholder: "{count, plural, one {Choose a primary tag} other {Choose {count} primary tags}}"
 ```
 
-Of course English has only two variants: singular or plural. You'll need to provide additional variants when creating translations for a language that has more than one plural form. If you need detailed information about the number of variants required for a language &mdash; or the order in which they should be listed &mdash; you can refer directly to the [pluralization rules](https://github.com/symfony/symfony/blob/3.3/src/Symfony/Component/Translation/PluralizationRules.php) that Flarum uses to map the variable to plural forms.
+Of course English has only two variants: singular or plural. You may want to provide additional variants when creating translations for a language that has more than one plural form.
+
+See the [Symfony Translation docs](https://symfony.com/doc/current/translation/message_format.html) for more complex examples.
```
SychO9, line 201:
```suggestion
See the [Symfony Translation docs](https://symfony.com/doc/5.2/translation/message_format.html) for more complex examples.
```

docs/extend/i18n.md
```diff
@@ -203,16 +208,11 @@ Support for grammatical gender will be added in a future version of Flarum. Deta
...
-```php
-$translator = app('translator');
-$translator = app('Flarum\Locale\Translator');
-$translator = app(Translator::class); // you'll need to import the class with 'use' for this
+First, you'll need to get an instance of the `Flarum\Locale\Translator` class. You'll usually do this by typehinting this class in the constructor of your class so a translator instance is injected by the [IoC Container](https://laravel.com/docs/8.x/container). If dependency injection is not available, you can use `resolve(Translator::class)`. Don't forget a `use Flarum\Locale\Translator;` statement at the top of your PHP file!
 ```
 
 Then, the API is similar to the JavaScript Translator class. You can use `$translator->trans` like you'd use `app.translator.trans` in JavaScript.
-You can learn more about the Translator's methods in [Symfony's `Translator` documentation](https://github.com/symfony/symfony/blob/3.3/src/Symfony/Component/Translation/Translator.php), which Flarum's `Translator` extends.
+You can learn more about the Translator's methods in [Symfony's `Translator` documentation](https://symfony.com/doc/current/translation/message_format.html), which Flarum's `Translator` extends.
```
SychO9, line 215:
```suggestion
You can learn more about the Translator's methods in [Symfony's `Translator` documentation](https://symfony.com/doc/5.2/translation/message_format.html), which Flarum's `Translator` extends.
```

docs/extend/permissions.md
```diff
@@ -45,6 +45,30 @@ Permissions are just part of the puzzle: if you're enforcing whether a user can
...
+
+The model prefix is useful in case you have different models but similar permissions (`flarum-sponsors.discussion.sponsor` vs `flarum-sponsors.post.sponsor`).
+
+### "Magic" model namespaces
+
+You may have seen some calls to `$actor->can` that don't use the full name of a permission; for example, `$actor->can('reply', $discussion)`, where the backing permission is actually called `discussion.reply`.
+
+This is done in core to make authorization calls shorter and simpler. Essentially, if the second argument is a discussion, Core's [DiscussionPolicy](https://github.com/flarum/core/blob/bba6485effc088e38e9ae0bc8f25528ecbee3a7b/src/Discussion/Access/DiscussionPolicy.php#L39-L44) will check the `discussion.PROVIDED_ABILITY` permission automatically.
+
+This can be used by extensions when a model namespace isn't present: for example, `$actor->can('someAbility, $discussion)` will check the `discussion.someAbility` permission if the $discussion argument is an instance of the `Discussion` model. However, this means you can't prefix your permissions with extension namespaces (or you have to put the extension namespace at the end).
```
SychO9, line 64:
```suggestion
This can be used by extensions when a model namespace isn't present: for example, `$actor->can('someAbility, $discussion)` will check the `discussion.someAbility` permission if the `$discussion` argument is an instance of the `Discussion` model. However, this means you can't prefix your permissions with extension namespaces (or you have to put the extension namespace at the end).
```

docs/extend/service-provider.md
```diff
@@ -29,22 +29,28 @@ A custom service provider should extend `Flarum\Foundation\AbstractServiceProvid
...
     {
         // custom logic here
     }
 }
 ```
 
-The `register` method will run during step (3) above, and the `boot` method will run during step (5) above. In both methods, the container is available via `$this->app`.
+The `register` method will run during step (3) above, and the `boot` method will run during step (5) above. In the `register` method, the container is available via `$this->container`. In the `boot` method, the container (or any other arguments), should be injected via typehinted method arguments.
+
+Flarum does not currently support Laravel Octane, but some [best practices](https://github.com/laravel/octane#dependency-injection--octane), like not injecting the container and using the `$container` argument inside `bind`, `singleton`, and `resolving` callbacks instead of `$this->container` should be used.
```
SychO9, line 53: Actually for singletons they recommend passing a callback that returns the container instance.

docs/extend/update-1.0.md
```diff
@@ -0,0 +1,261 @@
...
+- Developers need to decide between using `trans` or `transChoice` for pluralization
+- The pluralization format is proprietary to Symfony
+- We have to maintain the JS port ourselves
+- Keys for values provided to backend translations need to be wrapped in curly braces. (e.g. `$this->translator->trans('some.translation', ['{username}' => 'Some Username!'])`).
+- There's no support for complex applications (nested pluralization, non-number-based selection)
+- As a result of the previous point, genderization is impossible. And that's kinda important for a lot of languages.
+
+### New System
+
+In v5, Symfony dropped their proprietary `transChoice` system in favor of the more-or-less standard [ICU MessageFormat](https://symfony.com/doc/current/translation/message_format.html). This solves pretty much every single one of the aforementioned issues. In this release, Flarum will be fully switching to ICU MessageFormat as well. What does this mean for extensions?
```
SychO9, line 28:
```suggestion
In v5, Symfony dropped their proprietary `transChoice` system in favor of the more-or-less standard [ICU MessageFormat](https://symfony.com/doc/5.2/translation/message_format.html). This solves pretty much every single one of the aforementioned issues. In this release, Flarum will be fully switching to ICU MessageFormat as well. What does this mean for extensions?
```

docs/extend/update-1.0.md
```diff
@@ -0,0 +1,261 @@
...
+
+- `transChoice` should not be used at all; instead, the variable passed for pluralization should be included in the data
```
SychO9, line 30:
```suggestion
- `transChoice` should not be used at all; instead, the variable passed for pluralization should be included in the data.
```

docs/extend/update-1.0.md
```diff
@@ -0,0 +1,261 @@
...
+- Keys for backend translations no longer need to be surrounded by curly braces.
+- Translations can now use the [`select` and `plural`](https://symfony.com/doc/current/translation/message_format.html) formatter syntaxes. For the `plural` formatter, the `offset` parameter and `#` magic variables are supported
```
SychO9, line 32:
```suggestion
- Translations can now use the [`select` and `plural`](https://symfony.com/doc/5.2/translation/message_format.html) formatter syntaxes. For the `plural` formatter, the `offset` parameter and `#` magic variables are supported
```

docs/extend/admin.md
```diff
@@ -124,13 +136,18 @@ app.initializers.add('interstellar', function(app) {
         icon: 'fas fa-rocket', // Font-Awesome Icon
         label: app.translator.trans('acme-interstellar.admin.permissions.fly_rockets_label'), // Permission Label
         permission: 'discussion.rocket_fly', // Actual permission name stored in database (and used when checking permission).
+        tagScoped: false, // Should it be possible to apply this permission on tags, not just globally?
```
clarkwinkelmann, line 139: I'm confused by the way this is phrased as a question, it looks like this was copied from unfinished code with a TODO.

Maybe "Disallow applying this permission individually on each tag"

It might still be confusing since the way it works is explained below the example...

I think it might be more logical to show an example with `tag:Scoped: true` and just explain why you might want to use `false` in the documentation below (as is already the case)

docs/extend/assets.md
```diff
@@ -0,0 +1,8 @@
+# Extension Assets
+
+Some extensions might want to include assets like images or JSON files in their source code (note that this is not the same as uploads, which would probably require a [filesystem disk](filesystem.md)).
+
+This is actually very easy to do. Just create an `assets` folder at the root of your extension, and place any asset files there.
+Flarum will then automatically copy those files to its own `assets` directory every time the extension is enabled or [`php flarum assets:publish`](../console.md) is executed.
+
+Those files are copied into Flarum's assets folder and the URL to these files can be retrieved like this from the javascript: `app.forum.attribute('baseUrl') + '/assets/extensions/EXTENSION_ID/pumpkin.jpg'`, where EXTENSION_ID is the internal name of the extension (based on the package name, without / or flarum-ext-) and pumpkin.jpg is an example filename.
```
clarkwinkelmann, line 8: Love the filename ahah. I wonder where this was copied from :stuck_out_tongue: 

We don't have a way to access the assets based on the actual disk, do we? If a user changes the assets disk as described in another file above, then this will break. Not sure if all of Flarum is impacted (logo, avatars, ... do they break or automatically switch if you change the disk public URL?)

docs/extend/console.md
```diff
@@ -39,8 +39,23 @@ return [
...
 
-:::
+
+```php
+use Flarum\Extend;
+use YourNamespace\Console\CustomCommand;
+use Illuminate\Console\Scheduling\Event;
+
+return [
+  // Other extenders
```
clarkwinkelmann, line 53: Don't we format PHP examples with 4 spaces?

docs/extend/i18n.md
```diff
@@ -19,7 +25,7 @@ As a rule, a Flarum site can only display translations corresponding to the lang
 3. As a last-ditch effort, it will look for a "generic" English translation of the output.
 4. If none of the above are available, it will give up and display a translation key.
 
-Since English translations could be the only thing standing between forum users and unsightly translation keys, we strongly recommend including a complete set of English resources with every extension. (You will also need to include English resources if you wish to list your extension on the Flarum Marketplace.)
+Since English translations could be the only thing standing between forum users and unsightly translation keys, we strongly recommend including a complete set of English resources with every extension. (You will also need to include English resources if you wish to list your extension on the Flarum Marketplace, when that launches). Generally, extension developers will include english translations in their extension, and other language translations in a [language pack](language-packs.md).
```
clarkwinkelmann, line 28:
```suggestion
Since English translations could be the only thing standing between forum users and unsightly translation keys, we strongly recommend including a complete set of English resources with every extension. (You will also need to include English resources if you wish to list your extension on the Flarum Marketplace, when that launches). Generally, extension developers will include English translations in their extension, and other language translations in a [language pack](language-packs.md).
```

davwheat (changes requested):

clarkwinkelmann (changes requested):

docs/extend/assets.md
askvortsov1: Just rerun `php flarum assets:publish`

docs/extend/console.md
askvortsov1: I think we need something like https://github.com/rposborne/prettier-markdown-codeblocks

docs/extend/assets.md
clarkwinkelmann: I was referring to `app.forum.attribute('baseUrl') + '/assets/extensions/EXTENSION_ID/pumpkin.jpg'` likely not working in case the assets folder is customized.

I see that for logo and favicon we use `$assetsFilesystem->url($assetPath)` in the backend and serialize it to the frontend https://github.com/flarum/core/blob/d64750b3eb7de28d82727105b69c213cae5befe7/src/Api/Serializer/ForumSerializer.php#L115-L135 this is probably what we should also suggest to extensions?

I know I'm the one who originally shared the snippet that uses `baseUrl`, but if we include that in the documentation it's important this will only work with the local assets filesystem.

docs/extend/i18n.md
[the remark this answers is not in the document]
clarkwinkelmann: Oops, I suggested the same thing, and I think @askvortsov1 merged my commit suggestion. I'll mark this one as resolved.

docs/extend/assets.md
askvortsov1: Oh, this is true... Yeah I'll update that, good point!

docs/extend/service-provider.md
askvortsov1: You mean instead of injecting the container?

SychO9: I think I might've misread what you meant, but either way they literally just removed the documentation from the github repo readme and moved it to laravel docs site: https://laravel.com/docs/8.x/octane

clarkwinkelmann (approved):

SychO9 (approved):

PR #344, opened 2021-06-23
Correct typos in frontend.md

corydozen:
`webpack.config.json` should be `webpack.config.js`
`dev-dependencies` should be `devDependencies`


davwheat:
Could you also update the devDependencies to replace the `flarum` package with `flarum-tsconfig` at version `^1.0.0`?

davwheat (approved):

PR #346, opened 2021-06-28
Fix unexpected codeblock in i18n

Cadiducho:
This is currently the docs, with an unexpected codeblock which also erases the title of other section.

image


davwheat (approved):

davwheat:
Thank you for the PR! :)

PR #350, opened 2021-07-08
Add information about `flarum-tsconfig` and JS Typings

davwheat:


datitisev:
Any thoughts about https://github.com/flarum/flarum-tsconfig/issues/1 ?

davwheat:
@datitisev I've added a `baseUrl` key to the ts-config in these docs and in the package README. I've not run into the issue you described, but that key does seem to be required by TS for `paths` to function. I'm guessing VS Coe just infers it from your setup.

datitisev (approved):

PR #356, opened 2021-07-14
Use cleaner array syntax in example `config.php`

davwheat:


SychO9:
I'd rather we don't change this, that's what the default generated config file looks like:
https://github.com/flarum/core/blob/7eea2476ca72c0b15ccc154f9f5c564b7a96cdfd/src/Install/Steps/StoreConfig.php

davwheat:
Ah that makes sense. I've never actually used the default config file that's generated, so I've never noticed.

What about just moving the `=>`'s to the same columns for readability?

SychO9:
> What about just moving the =>'s to the same columns for readability?

Depends, is it more readable if it's the same or is it more readable if it's organised in the docs 🤔 ? I assume they would be in that page to understand and edit their own config file, so it might be easier for the eye if it's the same ? idk

luceos:
I would keep it the same and saving the config using short array cannot be done using raw PHP sadly.

PR #377, opened 2021-10-02
Document Flarum CLI

SychO9:
At first I only changed the references to fof extension generator and added a `cli.md` but I don't know if we need a dedicated docs page, instead I added references to commands in each section, similar to how Laravel's docs mention the appropriate artisan commands.


askvortsov1 (approved):
Great, thanks! At some point I think we should have a dedicated page / section that explains it a bit, but I like the commands being distributed across different pages. We need tutorial pages explaining how to do basic extensions, and those should use flarum CLI.

SychO9:
Sounds good!

PR #385, opened 2021-10-15
Update composer.md

datlechin:


clarkwinkelmann:
What's the reason for the change? I assume `jsonc` was used specifically to support JSON with comments. It seems like GitHub preview doesn't support it though.

datlechin:
`jsonc` is not supported.

Screenshot_20211016-093930_Chrome

askvortsov1:
> `jsonc` is not supported.
> 
> Screenshot_20211016-093930_Chrome
> 

As Matt suggested, let's use `json5` instead of `json`. Also, are there other parts of the code that need to be changed?

PR #398, opened 2021-12-02
Update folder ownership docs

askvortsov1:
Expand a bit on why this is necessary, add info on SELinux configuration.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -87,18 +87,18 @@ The user that Flarum is running as MUST have read + write access to:
 
 Extensions might require other directories, so you might want to recursively grant write access to the entire Flarum root install directory.
 
-There are several commands you'll need to run in order to set up file permissions.
+There are several commands you'll need to run in order to set up file permissions. Please note that if your install doesn't show warnings after executing just some of these, you don't need to run the rest.
 
 First, you'll need to allow write access to the directory. On Linux:
 
 ```bash
-sudo chmod 775 -R /path/to/directory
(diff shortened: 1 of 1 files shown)


docs/install.md
```diff
@@ -77,21 +77,37 @@ www.example.com {
...
-chmod 775 /path/to/directory
+chmod 775 -R /path/to/directory
 ```
 
-If Flarum requests write access to both the directory and its contents, you need to add the `-R` flag so that the permissions are updated for all the files and folders within the directory:
+If that isn't enough, you may need to check that your files are owned by the correct group and user. By default, in most Linux distributions `www-data` is the group and user that both PHP and the web server operate under. You'll need to look into the specifics of your distro and web server setup to make sure. You can change the folder ownership in most Linux operating systems by running:
 
 ```bash
-chmod 775 -R /path/to/directory
+chown -R www-data:www-data /path/to/directory
```
SychO9, line 101: Interesting thing to note here is that, my experience with ownership has been that the web user should be the group owner, and the actual user (cli) should be the owner, but that might only be applicable with the `package-manger` to still be able to run manual composer commands, and in-app composer commands, I can't actually think of other examples not involving that atm.

askvortsov1: The way I usually get around this is to add the CLI user to the web group. But that's a good thought.

SychO9 (approved):

tankerkiller125 (approved):

PR #399, opened 2021-12-11
docs: use Illuminate Str support func

davwheat:
`snake_case()`  doesn't seem defined when you try and use it. This seems like a better solution.


tankerkiller125 (approved):
It's not defined anymore because I think we removed all the helpers sometime in beta. This was most likely something we missed in the docs after.

askvortsov1 (approved):

PR #400, opened 2021-12-31
Document conventional commits specification & tweak merging process considerations.

SychO9:
**Changes Suggested**
- Officially switch to using the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#specification) specification, this will make changelog generation a lot easier through automation, but also makes the commit history more readable.
  - With the one custom difference being, describing the actual issue instead of the fix, when it comes to fix commits, as suggested by Daniël, this is better for the changelog.
- Tweak the merging process considerations
  - I don't believe we want regressions added to the milestone/project board.
  - I also believe we've taken the habit of creating issues in `core` that reference PRs in extensions that are added to the milestone. Although this won't matter once we switch to monorepo.

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -96,7 +96,8 @@ A typical contribution workflow looks like this:
 
 4. 💾 **Commit** your code with a descriptive message.
     * If your change resolves an existing issue (usually, it should) include "Fixes #123" on a newline, where 123 is the issue number.
-    * Write a [good commit message](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).
+    * Follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) specification.
+    * *Fix* commits should describe the issue fixed, not the fix.
 
 5. 🎁 **Submit** a Pull Request on GitHub.
     * Fill out the pull request template.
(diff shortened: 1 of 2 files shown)


SychO9:
btw, I don't believe the docs have been getting auto-updated with each new commit, the current docs don't have the latest commit changes in master.

davwheat (approved):

docs/contributing.md
```diff
@@ -96,7 +96,8 @@ A typical contribution workflow looks like this:
 
 4. 💾 **Commit** your code with a descriptive message.
     * If your change resolves an existing issue (usually, it should) include "Fixes #123" on a newline, where 123 is the issue number.
-    * Write a [good commit message](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).
+    * Follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) specification.
+    * *Fix* commits should describe the issue fixed, not the fix.
```
davwheat, line 100: I prefer this wording, but it's fine either way.

```suggestion
    * *Fix* commits should describe the issue fixed, not how it was fixed.
```

SychO9: much better indeed!

askvortsov1 (approved):

PR #410, opened 2022-01-23
[en] Add missing i18n code & Delete duplicate url

littlegolden:
1. There is no `contribute` Category i18n code in English json file, so neither does Crowdin.

2. A duplicate url of `Composer Introduction` needs to be deleted.


askvortsov1 (approved):
Thanks!

Issue #44, opened 2018-12-04
Documentation for API use


datitisev:
Even if it's as much as there previously was on the docs for beta 7, at least that'd be something which would include getting an API key and performing basic requests.

matteocontrini:
For anyone looking for the beta 7 JSON API documentation page, an archived version is available [on Archive.org](https://web.archive.org/web/20170727001822/http://flarum.org:80/docs/api/).

datitisev:
@matteocontrini It can also be seen at https://github.com/flarum/flarum.github.io/blob/20322c0e6011e4f304ae7e95f41594a0b086bc27/_docs/api.md.

However, with the new documentation for the API, we'd like to provide better information on each endpoint, parameters required, etc... as the old docs only included the endpoints.

jordanjay29:
Some relevant discussion in this (now closed) PR https://github.com/flarum/docs/pull/76

askvortsov1:
Without getting into specific endpoints, I'm planning to add a page describing general use of the REST API (most importantly, how to authenticate and deal with CSRF stuff.

askvortsov1:
https://github.com/zircote/swagger-php might be exactly what we need to generate the docs from code.

SychO9:
where would we want to have the docblocks for each endpoint, in the routes file or maybe on each controller class file ?

askvortsov1:
I would prefer the controller files; that way, changes to the API would prompt changes to the docs, and the file would still be readable (if we add it in routes, 95% of the file would be docblocks). That being said, my opinion on this isn't strong at all.

tankerkiller125:
Although that library looks great it relies on OpenAPI (Which is awesome). OpenAPI has issues in terms of documenting JSON-APIs properly, it can be done but the ones I've seen still usually end up missing something or being inaccurate in someway. With that said it would be better than what we currently have (no docs)

imorland:
Closing as stale — this issue is many years old (predates the 2.x docs and in most cases the current docs structure entirely). The docs have been substantially rewritten since. If this is still a real gap on the current site, please open a fresh issue with up-to-date detail.

PR #414, opened 2022-02-21
REST API documentation

clarkwinkelmann:
Fixes part of #44

While we still research a solution for documenting endpoints, the absence of documentation on the authentication and error management is really lacking, so I went ahead and made a first version.

### DIFF
--- a/docs/rest-api.md
+++ b/docs/rest-api.md
@@ -1,8 +1,14 @@
-# REST API
+# Consuming the REST API
 
 Flarum exposes a REST API which is used by the single page application but also available for external scripts.
 
 The API follows the best practices defined by the [JSON:API](https://jsonapi.org/) specification.
+
+:::info
+
+To extend the REST API with new endpoints, see [API and Data Flow](extend/api.md) in the developer documentation.
+
+:::
 
 ## Authentication
 
--- a/docs/rest-api.md
+++ b/docs/rest-api.md
@@ -1,0 +1,383 @@
+# REST API
+
+Flarum exposes a REST API which is used by the single page application but also available for external scripts.
+
+The API follows the best practices defined by the [JSON:API](https://jsonapi.org/) specification.
+
+## Authentication
+
+The single page app uses session cookies to authenticate against the API.
+External scripts can use stateless authentication using [API Keys](#api-keys) or [Access Tokens](#access-tokens).
+
+`GET` endpoints can be used without authentication.
+Only content visible to guests will be returned.
+Other endpoints generally cannot be used without authentication because of the [CSRF protection](#csrf-protection).
+
+### API keys
+
+API Keys are the primary way for scripts, tools and integrations to interact with Flarum.
+
+#### Creation
+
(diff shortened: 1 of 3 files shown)


docs/rest-api.md
```diff
@@ -0,0 +1,383 @@
...
+- `scopes`: Not implemented.
+- `created_at`: Can be set to any date, but is meant for the date of creation of the key.
+- `last_activity_at`: Will be updated automatically when the token is used.
+
+#### Usage
+
+Attach your key value to each API request using the `Authorization` header.
+Then provide the user ID you want to interact as at the end of the header:
+
+    Authorization: Token yourApiKeyValue; userId=1
```
askvortsov1, line 42:
```suggestion
    Authorization: Token YOUR_API_KEY_VALUE; userId=1
```

Just a nitpick, but this format might make it easier to see that this should be replaced by the value.

sidebars.js
```diff
@@ -56,6 +56,7 @@ module.exports = {
       label: 'Advanced',
       collapsible: false,
       items: [
+        'rest-api',
```
askvortsov1, line 59: I wonder if it might make sense to add a link to this page (not a docs reference but an absolute link) to the "extend" sidebar as well. It's not quite an extension, but some developers might look there anyway.

askvortsov1 (approved):

sidebars.js
clarkwinkelmann: Yeah I wasn't sure about this.

It would make sense for a "REST API" page to exist on both regular documentation and extension documentation, where one explains how to use the API and the other explains how to extend the API.

But at the same time extensions will also use the existing REST API so it does make sense to cross-link.

Maybe the page should be titled more clearly, something like "Consuming REST API consumption" so when there's an eventual "Extending REST API" it's more clear which one is meant for who?

docs/rest-api.md
clarkwinkelmann: Makes sense. I'll see if there are other comments before batch-committing the changes.

sidebars.js
tankerkiller125: I agree that clearing it up with "Consuming" and "Extending" would be very helpful in the long run.

clarkwinkelmann: Should I change just the title or also the URL?

I suppose the URL structure will change later on if we introduce a proper endpoint documentation, and we might move/refactor this document into the new pages.

clarkwinkelmann: Ok so there is already a "API and Data Flow" page that handles this subject in the Extend section, so I don't think it makes sense to add more pages or links in the sidebar. Maybe rename the name or URL of `extend/api.md` ?

For now I have added links between the 2 pages via info blocks.

clarkwinkelmann:
I have made changes based on the earlier comments. Anything else needed to get this merged?

askvortsov1 (approved):
LGTM! Thank you for taking this on. @tankerkiller125 anything on your end before this is merged?

tankerkiller125 (approved):
I'm good with this.

PR #418, opened 2022-03-12
Monorepo changes & Package Manager Internal Docs

SychO9:
* Moves package manager docs from the wiki of the repo to the internal docs.
* Tweaks the contributing guide to point to the monorepo.
* Update core GitHub code links to point to new monorepo.


SychO9:
Didn't change extension code links, they're probably more dev friendly than the monorepo, and unlike core, all links should still work with extensions even if we move them at some point.

docs/contributing.md
```diff
@@ -117,7 +117,7 @@ Don't worry if your code styling isn't perfect! StyleCI and Prettier will automa
...
+Flarum follows the [PSR-2](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md) coding standard and the [PSR-4](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-4-autoloader.md) autoloading standard. On top of this, we conform to a number of [other style rules](https://github.com/flarum/framework/blob/master/.styleci.yml). We use PHP 7 type hinting and return type declarations where possible, and [PHPDoc](https://docs.phpdoc.org/) to provide inline documentation. Try and mimic the style used by the rest of the codebase in your contributions.
```
askvortsov1, line 120: This should probably use the `main` branch

docs/contributing.md
```diff
@@ -28,16 +28,15 @@ Since Flarum is so extension-driven, we highly recommend [our extension docs](ex
...
 ```bash
 git clone https://github.com/flarum/flarum.git
 cd flarum
 
-# Set up a Composer path repository for Flarum packages
-composer config repositories.0 path "packages/*"
-git clone https://github.com/<username>/core.git packages/core
-git clone https://github.com/<username>/tags.git packages/tags # etc
+# Set up a Composer path repository for Flarum monorepo packages
+composer config repositories.0 path "packages/framework/*"
```
askvortsov1, line 38:
```suggestion
# Set up a Composer path repository for Flarum monorepo packages
composer config repositories.0 path "PATH_TO_MONOREPO/*/*"
```

askvortsov1: Note that the double star (/*/*) is deliberate. Should we mention that?

docs/contributing.md
```diff
@@ -28,16 +28,15 @@ Since Flarum is so extension-driven, we highly recommend [our extension docs](ex
 
 ### Setting Up a Local Codebase
 
-[flarum/flarum](https://github.com/flarum/flarum) is a "skeleton" application which uses Composer to download [flarum/core](https://github.com/flarum/core) and a [bunch of extensions](https://github.com/flarum). In order to work on these, we recommend forking and cloning them into a [Composer path repository](https://getcomposer.org/doc/05-repositories.md#path):
+[flarum/flarum](https://github.com/flarum/flarum) is a "skeleton" application which uses Composer to download the core package and a bunch of extensions within the monorepo [flarum/framework](https://github.com/flarum/framework). In order to work on these, we recommend forking and cloning the repository into a [Composer path repository](https://getcomposer.org/doc/05-repositories.md#path):
```
askvortsov1, line 31:
```suggestion
[flarum/flarum](https://github.com/flarum/flarum) is a "skeleton" application which uses Composer to download the core package and a bunch of extensions. Source code for Flarum core, extensions, and all packages used by the aforementioned is located in the Flarum monorepo [flarum/framework](https://github.com/flarum/framework). In order to contribute to these, you'll need to fork and clone the monorepo repository locally, and then add it to your dev environment as a [Composer path repository](https://getcomposer.org/doc/05-repositories.md#path):
```

askvortsov1 (changes requested):

docs/contributing.md
SychO9:
```suggestion
Flarum follows the [PSR-2](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md) coding standard and the [PSR-4](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-4-autoloader.md) autoloading standard. On top of this, we conform to a number of [other style rules](https://github.com/flarum/framework/blob/main/.styleci.yml). We use PHP 7 type hinting and return type declarations where possible, and [PHPDoc](https://docs.phpdoc.org/) to provide inline documentation. Try and mimic the style used by the rest of the codebase in your contributions.
```

docs/contributing.md
SychO9: Oh right, we changed the monorepo structure, I don't believe a mention is necessary, most will copy past.

docs/contributing.md
```diff
@@ -49,17 +48,17 @@ After your local installation is set up, make sure you've enabled `debug` mode i
...
 The process is the same for extensions.
 
 ```bash
-cd packages/tags/js
+cd packages/framework/extensions/tags/js
 npm install
-npm link ../../core/js
+npm link ../../../../framework/core/js
 npm run dev
 ```
```
SychO9, line 63: I missed this unnecessary bit. Yarn workspaces for the win, I should go through mentions of `npm` and change them to `yarn` as that's what we use now in the monorepo.

docs/contributing.md
```diff
@@ -28,16 +28,15 @@ Since Flarum is so extension-driven, we highly recommend [our extension docs](ex
...
 git clone https://github.com/flarum/flarum.git
 cd flarum
 
-# Set up a Composer path repository for Flarum packages
-composer config repositories.0 path "packages/*"
-git clone https://github.com/<username>/core.git packages/core
-git clone https://github.com/<username>/tags.git packages/tags # etc
+# Set up a Composer path repository for Flarum monorepo packages
+composer config repositories.0 path "PATH_TO_MONOREPO/*/*"
+git clone https://github.com/<username>/framework.git packages/framework
```
SychO9, line 39: I think it's easier to have it as `packages` directly

```suggestion
composer config repositories.0 path "packages/*/*"
git clone https://github.com/<username>/framework.git packages/framework
```

askvortsov1: Not sure about that, actually. `packages` could refer to the new monorepo, or to a generic packages directory where all flarum extensions are installed.

docs/contributing.md
askvortsov1: I am very happy with this change :grin:

askvortsov1 (approved):

docs/contributing.md
SychO9: I was more referring to the fact that the command right after it clones the monorepo fork to the same packages path, so it's a no brainer copy past for new contributors.

Though I don't feel strongly about this, just a thought I had when I noticed the second command.

Will make them consistent.

askvortsov1: Ah but then it doesn't work, as `packages/*/*` would refer to the folders in the root level of `framework`, not the extensions/packages inside.

If we want to be very general, perhaps we should clone to `monorepos/flarum`, and then have the path be `monorepos/*/*/*`?

SychO9: Oh wait, I missed what you actually meant. True, it wouldn't be correctly pointing to the actual monorepo packages.

perhaps a simple change to `packages/framework/*/*`, even though this wouldn't take into consideration local 3rd party extensions, since it's not the point of the contributing guide I think it should be fine, for extension development we have this https://docs.flarum.org/extend/start#extension-packaging.

For the more general solution, I don't think there'll be many monorepos in the ecosystem, and even if there will be, they probably won't have the same subdirectory structure Flarum has, so not sure it's worth it.

SychO9: Let's just leave it as `PATH_TO_MONOREPO` 👍🏼, both commands are in sync now.

PR #422, opened 2022-04-20
Fix Caddy example

francislavoie:
Request matchers in Caddy are exact, so `/assets` would only match requests to exactly `/assets` but not `/assets/app.js` etc. Using the `*` is necessary to match all requests to files within that directory. See https://caddyserver.com/docs/caddyfile/matchers#path-matchers


davwheat:
Would it be better to do `/assets/*`? I worry that if an extension added a page named `/assets-xxx` that it would then be inadvertently covered by this rule.

francislavoie:
Yeah, fair enough

davwheat (approved):

PR #433, opened 2022-07-27
Update

Justman10000:


docs/contributing.md
```diff
@@ -34,6 +34,10 @@ Since Flarum is so extension-driven, we highly recommend [our extension docs](ex
 git clone https://github.com/flarum/flarum.git
 cd flarum
 
+# Or, when you want to clone in the actually directory
```
davwheat, line 37:
```suggestion
# Or, when you want to clone directly into the current directory
```

docs/contributing.md
```diff
@@ -51,7 +55,7 @@ To contribute to the frontend, first install the JavaScript dependencies. The mo
 
 ```bash
 cd packages/framework
-yarn install
+yarn
```
davwheat, line 58: I think it's better to recommend using the full command, even though yarn on its own does the same thing.

```suggestion
yarn install
```

docs/code-of-conduct.md
```diff
@@ -29,7 +29,7 @@ When you're sure you're ready to start a discussion, please keep the following p
 - Give it a good title! You'll get the best results if your title makes it clear what you want to talk about.
 - Choose the right tag(s). This will increase the likelihood your post will be read and answered promptly.
 - *Don’t* post repeatedly about the same topic, as doing so will tend to have the opposite effect. 
-- If not using a tag set aside for multilingual use, *post in English only.* We can't help if we can't read you!
+- If not using a tag set aside for multilingual use, *post in English only.* We can't help you if we don't understand your posts
```
luceos, line 32:
```suggestion
- If not using a tag set aside for multilingual use, *post in English only.* We can't help you if we don't understand your posts.
```

docs/contributing.md
```diff
@@ -164,4 +168,4 @@ You confirm that you are able to grant us these rights. You represent that You a
 
 You represent that the Contributions are Your original works of authorship, and to Your knowledge, no other person claims, or has the right to claim, any right in any invention or patent related to the Contributions. You also represent that You are not legally obligated, whether by entering into an agreement or otherwise, in any way that conflicts with the terms of this license.
 
-The Flarum Foundation acknowledges that, except as explicitly described in this Agreement, any Contribution which you provide is on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING, WITHOUT LIMITATION, ANY WARRANTIES OR CONDITIONS OF TITLE, NON-INFRINGEMENT, MERCHANTABILITY, OR FITNESS FOR A PARTICULAR PURPOSE.
+The Flarum Foundation acknowledges that, except as explicitly described in this Agreement, any Contribution which you provide is on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING, WITHOUT LIMITATION, ANY WARRANTIES OR CONDITIONS OF TITLE, NON-INFRINGEMENT, MERCHANTABILITY, OR FITNESS FOR A PARTICULAR PURPOSE.
```
luceos, line 171: What is the change here? This seems to be part of the MIT license and it shouldn't be modified.

luceos (changes requested):

docs/contributing.md
davwheat: New line at the end of the file.

This matches with what we have in every other repository.

luceos (approved):

PR #441, opened 2022-11-21
Document PHPStan usage in extensions

SychO9:
Goes hand-in-hand with: https://github.com/flarum/framework/pull/3666
This also needs updating the Flarum CLI before merging.

### DIFF
--- a/docs/extend/models.md
+++ b/docs/extend/models.md
@@ -137,20 +137,19 @@ With all your snazzy new database tables and columns, you're going to want a way
 
 If you've added a new table, you'll need to set up a new model for it. Rather than extending the Eloquent `Model` class directly, you should extend `Flarum\Database\AbstractModel` which provides a bit of extra functionality to allow your models to be extended by other extensions. See the Eloquent docs linked above for examples of what your model class should look like.
 
-<!--
 ### Extending Models
 
 If you've added columns to existing tables, they will be accessible on existing models. For example, you can grab data from the `users` table via the `Flarum\User\User` model.
 
-If you need to define any attribute [accessors](https://laravel.com/docs/8.x/eloquent-mutators#defining-an-accessor), [mutators](https://laravel.com/docs/8.x/eloquent-mutators#defining-a-mutator), [dates](https://laravel.com/docs/8.x/eloquent-mutators#date-mutators), [casts](https://laravel.com/docs/8.x/eloquent-mutators#attribute-casting), or [default values](https://laravel.com/docs/8.x/eloquent#default-attribute-values) on an existing model, you can use the `Model` extender:
+<!-- If you need to define any attribute [accessors](https://laravel.com/docs/8.x/eloquent-mutators#defining-an-accessor), [mutators](https://laravel.com/docs/8.x/eloquent-mutators#defining-a-mutator), [dates](https://laravel.com/docs/8.x/eloquent-mutators#date-mutators), [casts](https://laravel.com/docs/8.x/eloquent-mutators#attribute-casting), or [default values](https://laravel.com/docs/8.x/eloquent#default-attribute-values) on an existing model, you can use the `Model` extender: 
 
 ```php
 use Flarum\Extend;
 use Flarum\User\User;
 
 return [
-    new Extend\Model(User::class)
-        ->defaultValue('is_alive', true)
+    (new Extend\Model(User::class))
+        ->default('is_alive', true)
         ->accessor('first_name', function ($value) {
             return ucfirst($value)
         })
@@ -162,6 +161,20 @@ return [
 ];
 ```
 -->
+
+If you need to define any attribute [casts](https://laravel.com/docs/8.x/eloquent-mutators#attribute-casting), or [default values](https://laravel.com/docs/8.x/eloquent#default-attribute-values) on an existing model, you can use the `Model` extender:
+
+```php
+use Flarum\Extend;
+use Flarum\User\User;
+
+return [
+    (new Extend\Model(User::class))
+        ->default('is_alive', true)
+        ->cast('suspended_until', 'datetime')
+        ->cast('is_admin', 'boolean')
+];
+```
 
 ### Relationships
 
--- a/docs/extend/static-code-analysis.md
+++ b/docs/extend/static-code-analysis.md
@@ -26,7 +26,7 @@ Next, you need to create a `phpstan.neon` file in the root of your extension. Th
 
 ```neon
 includes:
-  - vendor/php-packages/phpstan/extension.neon
+  - vendor/flarum/phpstan/extension.neon
 
 parameters:
 	# The level will be increased in Flarum 2.0
@@ -37,7 +37,7 @@ parameters:
 	excludePaths:
 		- *.blade.php
 	checkMissingIterableValueType: false
-	databaseMigrationsPath: ['framework/core/migrations']
+	databaseMigrationsPath: ['migrations']
 ```
 
 Finally, you need to add the following script to your `composer.json` file:
--- a/docs/extend/static-code-analysis.md
+++ b/docs/extend/static-code-analysis.md
@@ -29,15 +29,15 @@ includes:
   - vendor/flarum/phpstan/extension.neon
 
 parameters:
-	# The level will be increased in Flarum 2.0
-	level: 5
-	paths:
-	  - src
-	  - extend.php
-	excludePaths:
-		- *.blade.php
-	checkMissingIterableValueType: false
-	databaseMigrationsPath: ['migrations']
+  # The level will be increased in Flarum 2.0
+  level: 5
+  paths:
+    - src
+    - extend.php
+  excludePaths:
+    - *.blade.php
+  checkMissingIterableValueType: false
+  databaseMigrationsPath: ['migrations']
 ```
 
 Finally, you need to add the following script to your `composer.json` file:
--- a/docs/extend/static-code-analysis.md
+++ b/docs/extend/static-code-analysis.md
@@ -1,0 +1,86 @@
+# Static Code Analysis
+
+Static code analysis is the process of analyzing the source code against a set of rules to find bugs, code smells, and security vulnerabilities. This is a great way to improve the quality of your code and to find potential issues before they are deployed to production. An example is validating the typings of a function to ensure that the function is called with the correct arguments.
+
+Flarum provides a static code analysis package based on [PHPStan](https://phpstan.org/) that can be added to your extension. In this guide, we will show you how to add the package to your extension and how to run the analysis.
+
+## Setup
+
+:::tip [Flarum CLI](https://github.com/flarum/cli)
+
+You can use the CLI to automatically add and update the infrastructure for phpstan to your code:
+
+```bash
+$ flarum-cli infra phpstan
+```
+
+:::
+
+First you need to require the `flarum/phpstan` package in your extension. You can do this by running the following command in the root of our extension:
+
+```bash
+composer require --dev flarum/phpstan:^1.0
+```
+
+Next, you need to create a `phpstan.neon` file in the root of your extension. This file contains [the configuration for PHPStan](https://phpstan.org/config-reference). You can copy the following configuration into the file:
+
+```neon
+includes:
+  - vendor/php-packages/phpstan/extension.neon
+
+parameters:
+	# The level will be increased in Flarum 2.0
+	level: 5
+	paths:
+	  - src
+	  - extend.php
+	excludePaths:
+		- *.blade.php
+	checkMissingIterableValueType: false
+	databaseMigrationsPath: ['framework/core/migrations']
+```
+
+Finally, you need to add the following script to your `composer.json` file:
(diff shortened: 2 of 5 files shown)


n-peugnet:
This is very cool, can't wait to use it!

It seems another thing missing, besides the CLI changes, is to publish the package to Packagist repository. Could you do it, even without tagged version? I would love to try this out.

docs/extend/models.md
```diff
@@ -137,20 +137,19 @@ With all your snazzy new database tables and columns, you're going to want a way
...
 
-<!--
 ### Extending Models
 
 If you've added columns to existing tables, they will be accessible on existing models. For example, you can grab data from the `users` table via the `Flarum\User\User` model.
 
-If you need to define any attribute [accessors](https://laravel.com/docs/8.x/eloquent-mutators#defining-an-accessor), [mutators](https://laravel.com/docs/8.x/eloquent-mutators#defining-a-mutator), [dates](https://laravel.com/docs/8.x/eloquent-mutators#date-mutators), [casts](https://laravel.com/docs/8.x/eloquent-mutators#attribute-casting), or [default values](https://laravel.com/docs/8.x/eloquent#default-attribute-values) on an existing model, you can use the `Model` extender:
+<!-- If you need to define any attribute [accessors](https://laravel.com/docs/8.x/eloquent-mutators#defining-an-accessor), [mutators](https://laravel.com/docs/8.x/eloquent-mutators#defining-a-mutator), [dates](https://laravel.com/docs/8.x/eloquent-mutators#date-mutators), [casts](https://laravel.com/docs/8.x/eloquent-mutators#attribute-casting), or [default values](https://laravel.com/docs/8.x/eloquent#default-attribute-values) on an existing model, you can use the `Model` extender:
```
askvortsov1, line 144: Doesn't this comment need an end?

askvortsov1: Why not remove this section in favor of the new one?

docs/extend/static-code-analysis.md
```diff
@@ -0,0 +1,86 @@
...
+{
+  "scripts": {
+    "analyse:phpstan": "phpstan analyse",
+    "clear-cache:phpstan": "phpstan clear-result-cache"
+  },
+  "scripts-descriptions": {
+    "analyse:phpstan": "Run static analysis"
+  }
+}
+```
```
askvortsov1, line 55: Shouldn't all of this be done by the CLI? If so, can we denote that you need to run the CLI command OR do this?

askvortsov1 (approved):
We can probably do that, and merge this, when we release 2.7

docs/extend/static-code-analysis.md
```diff
@@ -0,0 +1,86 @@
...
+
+```bash
+composer require --dev flarum/phpstan:^1.0
+```
+
+Next, you need to create a `phpstan.neon` file in the root of your extension. This file contains [the configuration for PHPStan](https://phpstan.org/config-reference). You can copy the following configuration into the file:
+
+```neon
+includes:
+  - vendor/php-packages/phpstan/extension.neon
```
n-peugnet, line 29: This should likely be `flarum/phpstan` for extension developers:

```suggestion
  - vendor/flarum/phpstan/extension.neon
```

docs/extend/static-code-analysis.md
```diff
@@ -0,0 +1,86 @@
...
+parameters:
+	# The level will be increased in Flarum 2.0
+	level: 5
+	paths:
+	  - src
+	  - extend.php
+	excludePaths:
+		- *.blade.php
+	checkMissingIterableValueType: false
+	databaseMigrationsPath: ['framework/core/migrations']
```
n-peugnet, line 40: This path also seems strange for out-of-tree extensions.

docs/extend/testing.md
```diff
@@ -10,9 +10,21 @@ It is essentially a collection of utils that allow testing Flarum core and exten
...
+$ flarum-cli infra backendTesting
+```
+
+:::
+
 Firstly, you will need to require the `flarum/testing` composer package as a dev dependency for your extension:
 
-`composer require --dev flarum/testing:^1.0`
+```bash
+composer require --dev flarum/testing:^1.0`
```
n-peugnet, line 26: Looks like this backtick should not stay here:

```suggestion
composer require --dev flarum/testing:^1.0
```

n-peugnet:
I just remembered that I can simply install the package from the git repo while it is not released yet. So I tested it and it seems there is a problem with the `stubFiles` that are included from the `vendor` directory.

First I had to set the `include` path as so as I said above https://github.com/flarum/docs/pull/441#discussion_r1097712669:

```neon
# phpstan.neon
includes:
  - vendor/flarum/phpstan/extension.neon
```

But then I had the following error:

    Note: Using configuration file /home/nicolas/Source/www/flarum/packages/cross-references/phpstan.neon.
    Stub file /home/nicolas/Source/www/flarum/packages/cross-references/vendor/vendor/nunomaduro/larastan/stubs/Enumerable.stub does not exist.

And indeed, the paths of the stub files are wrong:

```neon
# vendor/flarum/phpstan/larastan-extension.neon
parameters:
    stubFiles:
        - ../../vendor/nunomaduro/larastan/stubs/Enumerable.stub
        - ../../vendor/nunomaduro/larastan/stubs/EloquentBuilder.stub
        - ../../vendor/nunomaduro/larastan/stubs/Collection.stub
        - ../../vendor/nunomaduro/larastan/stubs/EloquentCollection.stub

        [...]
```

This is because the way its used internally in Flarum source tree is one less level deep than what it will be once installed through composer.

One easy way I see to fix this without having to move the directory of the extension in `flarum/framwork` is to use the [`%rootDir%` path expansion from PHPStan](https://phpstan.org/config-reference#expanding-paths) (for instance `%rootDir%/../../vendor/nunomaduro/larastan/stubs/Enumerable.stub`).

n-peugnet:
I also have the following error messages:

```
Ignored error pattern #Relation '[A-z_-]+' is not found in [A-z ]+ model.# was not matched in reported errors.                                            
Ignored error pattern #^Parameter \#1 \$query of method [A-z_<>\\]+\:\:union\(\) expects [A-z_<> .,|\\]+ given\.$# was not matched in reported errors.    
Ignored error pattern #^Parameter \#1 \$query of method [A-z_<>\\]+\:\:joinSub\(\) expects [A-z_<> .,|\\]+ given\.$# was not matched in reported errors.  
Ignored error pattern #Template type T of function resolve[()]{2} is not referenced in a parameter.# was not matched in reported errors.                  
Ignored error pattern #^Call to an undefined method Illuminate\\Database\\ConnectionInterface\:\:[A-z0-9_]+\(\)\.$# was not matched in reported errors.   
```

This comes from the `phpstan-baseline.neon`:

```neon
# vendor/flarum/phpstan/phpstan-baseline.neon
parameters:
	ignoreErrors:
		# Remove this group below with larastan 2.0 (i.e Flarum 2.0)
		- "#Relation '[A-z_-]+' is not found in [A-z\_]+ model.#"
		- '#^Parameter \#1 \$query of method [A-z_<>\\]+\:\:union\(\) expects [A-z_<> .,|\\]+ given\.$#'
		- '#^Parameter \#1 \$query of method [A-z_<>\\]+\:\:joinSub\(\) expects [A-z_<> .,|\\]+ given\.$#'

		#	We ignore this because resolve can either take a class name as the generic return type or just a binding name.
		- "#Template type T of function resolve[()]{2} is not referenced in a parameter.#"

		#	We ignore new static errors because we want extensibility.
		#	@TODO: needs discussion.
		- "#^Unsafe usage of new static[()]{2}.$#"

		# ConnectionInterface lacks methods that exist in the implementation,
		# yet we don't want to inject the implementation.
		- '#^Call to an undefined method Illuminate\\Database\\ConnectionInterface\:\:[A-z0-9_]+\(\)\.$#'

```

It might be a good thing to [selectively disable reporting unused ignore](https://phpstan.org/user-guide/ignoring-errors#reporting-unused-ignores) for each of these ignoreErrors like this:

```neon
parameters:
	ignoreErrors:
		# Remove this group below with larastan 2.0 (i.e Flarum 2.0)
		- message: "#Relation '[A-z_-]+' is not found in [A-z\_]+ model.#"
		  reportUnmatched: false

		[...]
```

docs/extend/models.md
SychO9: cause it references a non existent `accessor` and `mutator` extenders that kind of act as a reminder to actually add them 😋

docs/extend/static-code-analysis.md
SychO9: When the CLI gets augmented to actually do this we will update this doc page ;)

docs/extend/static-code-analysis.md
SychO9: good catch!
```suggestion
	databaseMigrationsPath: ['migrations']
```

docs/extend/static-code-analysis.md
SychO9: actually, I already added a future command reference on the top of the page

n-peugnet (approved):

docs/extend/static-code-analysis.md
```diff
@@ -0,0 +1,86 @@
...
+### Extended model attribute types
+
+PHPStan needs to be able to determine the type of an attribute added to an existing model. To do this you can use the `Extend\Model(...)->cast(...)` method.
+
+For example, if your extension were to add a `is_cool` attribute to the `User` model, you can use [attribute casting](https://laravel.com/docs/8.x/eloquent-mutators#attribute-casting) to explicitly define the attribute as boolean. The `flarum/phpstan` package will automatically detect this and communicate it to PHPStan.
+
+```php
+(new Extend\Model(User::class))
+    ->cast('is_cool', 'bool'),
+```
```
n-peugnet, line 86: Just one question about this section. The last error that I couldn't fix is:

    Call to an undefined method Flarum\Discussion\Discussion::sources()

This is a [method I added in extend.php](https://github.com/club-1/flarum-ext-cross-references/blob/2731136be37dcdbabbd41846b6007d2d097a2a25/extend.php#L48-L50) like this:

```php
    (new Extend\Model(Discussion::class))
        ->belongsToMany('sources', Discussion::class, 'discussion_reference', 'target_id', 'source_id')
        ->belongsToMany('targets', Discussion::class, 'discussion_reference', 'source_id', 'target_id'),
```

How can I make PHPStan understand method declared like this?

I thought about using the `cast()` method as it seems it is supposed to solve this kind of issues but it looks like is is only for primitive types.

docs/extend/static-code-analysis.md
```diff
@@ -0,0 +1,86 @@
...
+parameters:
+	# The level will be increased in Flarum 2.0
+	level: 5
+	paths:
+	  - src
+	  - extend.php
+	excludePaths:
+		- *.blade.php
+	checkMissingIterableValueType: false
+	databaseMigrationsPath: ['migrations']
```
n-peugnet, line 40: Always use 2 space to keep indentation consistent in the example neon file:

```suggestion
includes:
  - vendor/flarum/phpstan/extension.neon

parameters:
  # The level will be increased in Flarum 2.0
  level: 5
  paths:
    - src
    - extend.php
  excludePaths:
    - *.blade.php
  checkMissingIterableValueType: false
  databaseMigrationsPath: ['migrations']
```

docs/extend/static-code-analysis.md
SychO9: The package should have logic for all relationship methods, I need to check again `belongsToMany` might have been fogotten

SychO9: @n-peugnet can you link to the codebase, `belongsToMany` is correctly picked up by the phpstan package when used on the framework, so not sure why it isn't working on your end.

n-peugnet: The extension is this one : https://github.com/club-1/flarum-ext-cross-references

I added phpstan setup in this branch : https://github.com/club-1/flarum-ext-cross-references/pull/25

For now it fails with this issue : https://github.com/club-1/flarum-ext-cross-references/actions/runs/4143686829/jobs/7165830767#step:4:15

    Stub file /home/runner/work/flarum-ext-cross-references/flarum-ext-cross-references/vendor/vendor/nunomaduro/larastan/stubs/Enumerable.stub does not exist.

As my commit is not yet available in [flarum/phpstan](https://github.com/flarum/phpstan).

SychO9: @n-peugnet it is now

n-peugnet: Thanks, I will push a new commit right now.

n-peugnet: @SychO9 I just updated `flarum/phpstan` to the latest version and now everything is working fine. The stub files are now found as expected and mysteriously, the `Call to an undefined method Flarum\Discussion\Discussion::sources()` error disappeared.

Well, good for us I guess!

SychO9: nice :) thanks for the help!

PR #458, opened 2023-04-30
docs: document config `unix_socket`

SychO9:


docs/config.md
```diff
@@ -25,6 +25,7 @@ Here's a quick overview of what everything means with an example file:
     'prefix' => '', // the prefix for the tables, useful if you are sharing the same database with another service
     'port' => '3306', // the port of the connection, defaults to 3306 with MySQL
     'strict' => false,
+    'unix_socket' => '', // the unix socket of the connection, useful if you are using a local database (example: '/run/mysqld/mysqld.sock')
```
askvortsov1, line 28: What happens if a `host` and `port` are both provided? I was thinking that it might be better to add a separate section at the bottom explaining how to use a unix_socket DB. Also, I don't believe the Flarum installer will allow setting up a site without a database, so we should document a workaround for installing a socket DB-based installation.

askvortsov1 (changes requested):

PR #459, opened 2023-05-24
Add Flarum scheduler guide to docs

iamdarkle:
This PR includes the Flarum scheduler guide (https://discuss.flarum.org/d/24118-setup-the-flarum-scheduler-using-cron) to the docs. I have included the guide under the Management section, I think it fits best (?).

I've taken the liberty to adjust a bit the content of Ian's original tutorial a bit, taking into account that this will be in the docs... Feel free to correct or change what you think is appropriate.


davwheat (changes requested):
We shouldn't use words like "easy" and "simple" because it might not be for some. Other than that, just some minor changes.

iamdarkle:
Much better now, thanks for the review :)

davwheat (approved):

PR #469, opened 2024-01-06
feat: package manager & installable packages

SychO9:
**putting here for use in discuss**
294670348-6e18c6ed-8abb-4feb-88a9-e8bb060e54ab


docs/extensions.md
```diff
@@ -32,8 +56,45 @@ Similarly to installation, to remove an extension:
...
+
+:::
+
+## Installing Non-stable extensions
+
+If for whatever reason you want to install a non-stable extension (e.g. a beta, alpha or RC version) you must first update the **Minimum stability** setting to the wanted stability.
+
+* If you set it to Alpha, you will be able to install alpha, beta, RC and stable versions.
+* If you set it to Beta, you will be able to install beta, RC and stable versions.
+* If you set it to RC, you will be able to install RC and stable versions.
```
luceos, line 99: RC = release candidate; either of these need to be fully written out

luceos:
The docs seem to imply the extension manager is already installed. Nowhere is mentioned how to install this manager in case it isn't installed, eg on existing systems. This might alleviate some support questions.

GreXXL:
> The docs seem to imply the extension manager is already installed. Nowhere is mentioned how to install this manager in case it isn't installed, eg on existing systems. This might alleviate some support questions.

In the downloaded package it's already bundled. I don't know if we also changed the skeleton when you manually install with composer so that it defaults installs as a core extension. In that case it would be fine to describe it "as already installed"?

SychO9:
I don't actually think we should bundle it for those installing from the CLI, but instead include the installation instructions of the extension manager i the section of installing flarum from the CLI, that way it's clear that if you wish to manage extensions from the interface (even though you are currently using the CLI to install) then here it is and how to add it.

Also need to add the extension's installation instructions in the extensions section like daniel mentions.

PR #466, opened 2023-10-21
2.0 docs

SychO9:
(*WIP*) here for a general view of changes so far.

PR #472, opened 2024-03-14
docs: JSON:API refactor

SychO9:
Documents breaking changes for the upgrade guide and links to the new documentation for API resources.

It adds links to tags extension from 1.x and the one from 2.x as an example for before and after.


docs/extend/api.md
```diff
@@ -12,348 +12,967 @@ To use the built-in REST API as part of an integration, see [Consuming the REST
...
+        Endpoint\Index::make()
+            ->defaultSort('-createdAt'),
+            ->paginate(),
+    ];
+}
+```
+
+#### Searching and Filtering
+
+Read our [searching and filtering](search.md) guide for more information!
```
askvortsov1, line 252: Unrelated to this PR, but I think docs.flarum.org might not be up to date

docs/extend/api.md
```diff
@@ -12,348 +12,967 @@ To use the built-in REST API as part of an integration, see [Consuming the REST
...
+    return [
+        Endpoint\Index::make()
+            ->defaultInclude(['parent']),
+    ];
+}
+```
+
+### Lifecycle Hooks
+
+Some methods on the endpoint allow you to hook certain logic into the lifecycle of the endpoint. These are `before`, `after`, and `beforeSerialization` which is often not very different from `after` but is called before it when available on the endpoint.
```
askvortsov1, line 420: Might be worth talking about what you might want to do here, and why you wouldn't do it in the database resource hooks.

docs/extend/api.md
```diff
@@ -12,348 +12,967 @@ To use the built-in REST API as part of an integration, see [Consuming the REST
...
+        Schema\DateTime::make('createdAt'),
+    ];
 }
 ```
 
-### Creating a Resource
+### Visibility
 
-For the controller that creates a resource, extend the `Flarum\Api\Controller\AbstractCreateController` class. This is the same as the show controller, except the response status code will automatically be set to `201 Created`. You can access the incoming JSON:API document body via `$request->getParsedBody()`:
+You can use the `visible` method to determine whether an attribute should be visible in the API response.
```
askvortsov1, line 561:
```suggestion
You can use the `visible` method to conditionally include an attribute in the API response.
```

docs/extend/api.md
```diff
@@ -12,348 +12,967 @@ To use the built-in REST API as part of an integration, see [Consuming the REST
...
 ```php
-use Flarum\Api\Controller\AbstractCreateController;
-use Illuminate\Support\Arr;
-use Psr\Http\Message\ServerRequestInterface as Request;
-use Tobscure\JsonApi\Document;
+use Flarum\Api\Context;
+use Flarum\Api\Schema;
 
-class CreateTagController extends AbstractCreateController
+public function fields(): array
```
askvortsov1, line 567: maybe worth noting somewhere towards the top that all these examples are of methods in the resource class.

docs/extend/api.md
```diff
@@ -12,348 +12,967 @@ To use the built-in REST API as part of an integration, see [Consuming the REST
...
+    return [
+        Schema\Str::make('name')
+            ->visible(fn (Label $label, Context $context) => $context->getActor()->can('edit', $label)),
+    ];
+}
+```
+
+### Writability
+
+By default, a field is not writable unless you specify it to be so. You can use the `writable` method to make a field writable.
```
askvortsov1, line 578: Can you specify custom logic for how a field will be written to the database?

EDIT: literally 2 sections below. Might be worth saying that it will write directly to the model's field by default, but will use a custom setter if provided?

docs/extend/api.md
```diff
@@ -12,348 +12,967 @@ To use the built-in REST API as part of an integration, see [Consuming the REST
...
+            }),
+    ];
 }
 ```
 
-### Including Relationships
+### Validation
 
-To include relationships when **listing**, **showing**, or **creating** your resource, specify them in the `$include` and `$optionalInclude` properties on your controller:
+You can use the `rule` method to add a [Laravel validation rule](https://laravel.com/docs/10.x/validation#available-validation-rules) to an attribute. Each type of attribute will have some custom validation rules available as methods.
```
askvortsov1, line 641:
```suggestion
You can use the `rule` method to add a [Laravel validation rule](https://laravel.com/docs/10.x/validation#available-validation-rules) to an attribute. We've provided helper methods on some attributes for common validation rules.
```

docs/extend/api.md
```diff
@@ -12,348 +12,967 @@ To use the built-in REST API as part of an integration, see [Consuming the REST
...
-        ->setLimit(20)
-        ->setMaxLimit(50)
-        ->setSort(['name' => 'asc'])
-        ->addSortField('firstName')
-        ->prepareDataQuery(function ($controller) {
-            // Add custom logic here to modify the controller
-            // before data queries are executed.
-        })
-]
+Integer::make('commentCount')
```
askvortsov1, line 707: Where would this code go?

docs/extend/api.md
```diff
@@ -12,348 +12,967 @@ To use the built-in REST API as part of an integration, see [Consuming the REST
...
+        Schema\Relationship\ToMany::make('children')
+            ->type('labels'),
+    ];
+}
 ```
 
-## Serializers
+### Inclusion & Linkage
+
+You can mark a relationship as includable through the `includable` method. This means that the relationship can be included in the API response. You can also use the `withLinkage` and `withoutLinkage` methods to determine whether the relationship ID(s) should be included in the API response (`ToMany` relationships are not linked by default contrary to `ToOne` relationships).
```
askvortsov1, line 745: When would you have non-includable relationships?

docs/extend/api.md
```diff
@@ -12,348 +12,967 @@ To use the built-in REST API as part of an integration, see [Consuming the REST
...
 
-Before we can send our data to the frontend, we need to convert it to JSON:API format so that it can be consumed by the frontend.
-You should become familiar with the [JSON:API specification](https://jsonapi.org/format/).
-Flarum's JSON:API layer is powered by the [tobscure/json-api](https://github.com/tobscure/json-api) library.
+:::danger
 
-A serializer is just a class that converts some data (usually [Eloquent models](models.md#backend-models)) into JSON:API.
-Serializers serve as intermediaries between backend and frontend models: see the [model documentation](models.md) for more information.
-To define a new resource type, create a new serializer class extending `Flarum\Api\Serializer\AbstractSerializer`. You must specify a resource `$type` and implement the `getDefaultAttributes` method which accepts the model instance as its only argument:
+Adding linkage for `ToMany` relationships can lead to performance issues as it will include the IDs of all the related models in the API response. This is why it is not linked by default.
```
askvortsov1, line 749: I think I don't fully understand linkage from this description. Is it a concept from Toby's server, or smething we added?

docs/extend/api.md
```diff
@@ -12,348 +12,967 @@ To use the built-in REST API as part of an integration, see [Consuming the REST
...
 
-    protected function getDefaultAttributes($discussion)
-    {
-        return [
-            'title' => $discussion->title,
-        ];
-    }
+### Polymorphic Relationships
+
+You use the `collection` method to define the resource types that a polymorphic relationship can point to.
```
askvortsov1, line 769: Are there docs by us or on the Eloquent site we can link to re: polymorphic relationships?

docs/extend/update-2_0.md
```diff
@@ -139,6 +165,11 @@ The search system has been refactored to allow for more flexibility and extensib
 * You can now get the total search result count from `SearchResults`.
 * You can now replace an existing filter implementation.
 
+### Miscellaneous
+
+##### <span class="breaking">Breaking</span>
+* The `(Extend\Notification)->type()` extender no longer accepts a serializer as second argument.
```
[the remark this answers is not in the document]
SychO9: nothing, before it was 3 args, now it's 2 args (without the serializer arg)

The new implementation can infer the API resource from the notification blueprint, since both blueprint and resource classes specify the model class.

static/en/img/api_flowchart.svg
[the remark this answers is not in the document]
SychO9: seems to only be an issue when opening the image, as long as it shows in the docs I think it's fine.

docs/extend/api.md
```diff
@@ -12,348 +12,967 @@ To use the built-in REST API as part of an integration, see [Consuming the REST
...
+
+    public function fields(): array
+    {
+        return [
+            Schema\Str::make('name')
+                ->requiredOnCreate()
+                ->writable(),
+            Schema\Str::make('description')
+                ->writable()
+                ->maxLength(700)
```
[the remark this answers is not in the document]
SychO9: good point

docs/extend/api.md
```diff
@@ -12,348 +12,967 @@ To use the built-in REST API as part of an integration, see [Consuming the REST
...
-```bash
-$ flarum-cli make backend api-controller
+```php
+use Flarum\Api\Endpoint;
+
+public function endpoints(): array
+{
+    return [
+        Endpoint\Index::make()
+            ->paginate(20, 50), // these are the default values, so you may omit them.
```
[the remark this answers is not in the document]
SychO9: just `->paginate()`, let me specifically say omit the arguments
```suggestion
            ->paginate(20, 50), // these are the default values, so you may omit these arguments.
```

docs/extend/api.md
SychO9: no, because that's 1.x, and this is for the 2.x branch which is still not published until 2.0 is done ;) https://github.com/flarum/docs/pull/466

docs/extend/api.md
SychO9: When you only need a relationship to be writable perhaps. Though not a common use case, it remains valid.

docs/extend/api.md
SychO9: Yes, it's not straightforward to know what it's referring to, perhaps an example would be good to have here.

Linkage is the inclusion of the relationship IDs, like:

```json
"attributes": { ... },
"relationships": { "firstPost": { "type": "posts", "id": 4 } }
```

docs/extend/api.md
SychO9: good point

docs/extend/api.md
```diff
@@ -12,348 +12,967 @@ To use the built-in REST API as part of an integration, see [Consuming the REST
...
+            SortColumn::make('createdAt'),
+        ];
+    }
+}
 ```
 
-:::caution
+### Resource Definition
+
+The API resource class must extend the `Flarum\Api\Resource\AbstractDatabaseResource` class when interacting with Eloquent models, and `Flarum\Api\Resource\AbstractResource` when not. The `type` method should return a unique string that identifies the resource type. In the case of a database resource, a model method must return the class name of the model.
```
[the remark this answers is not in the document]
SychO9: Not a very common way of referring to the class name 😬, how about

```suggestion
The API resource class must extend the `Flarum\Api\Resource\AbstractDatabaseResource` class when interacting with Eloquent models, and `Flarum\Api\Resource\AbstractResource` when not. The `type` method should return a unique string that identifies the resource type. In the case of a database resource, the `model` method must return the class name of the model (`::class` property).
```

docs/extend/api.md
SychO9: added tip explaining linkage in this section

docs/extend/api.md
askvortsov1: Might be worth adding a short tip about this

askvortsov1 (approved):

PR #479, opened 2024-11-03
docs: apply versioning (/1.x, /next)

SychO9:
https://docusaurus.io/docs/versioning#tagging-a-new-version

Applies the 1.x version so that we can now have 2.x docs content separately.

* url `/docs`: will continue to point to 1.x content until 2.0 stable
* url `/docs/next`: will point to 2.x docs
* url `/docs/1.x`: when 2.0 hits stable

Directory/editing wise, this means:
* dir `/docs/..`: is 2.x documentation
* dir `/versioned_content/version-1.x`: is 1.x content

Same for translations:
* `i18n/it/docusaurus-plugin-content-docs/current`: 2.x docs
* `i18n/it/docusaurus-plugin-content-docs/version-1.x`: 1.x docs


luceos:
Personally I wouldn't use `next`. Once 2.x is stable all url's change as it's no longer next. I would use 2.x for that structuring and then maybe give it higher precedence when released. Or mark it as "dev" or "nightly" somehow.

SychO9:
@luceos I looked into it, so we can have /2.x be a permenant path without changes, however the root path will remain pointing at 1.x, which I think is fine and if we decide to enforce a 1.x path it will just have to be done in addition to a server redirect rule

luceos:
Okay that makes sense.

luceos (approved):

PR #487, opened 2024-12-21
install.md: improved docs for "customizing paths" section

Ser5:
More structured and detailed text:
- at the top are general description of Flarum config files that will be used further
- then common cases of custom paths

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -162,25 +162,111 @@ By default Flarum's directory structure includes a `public` directory which cont
 
 However, if you wish to host Flarum in a subdirectory (like `yoursite.com/forum`), or if your host doesn't give you control over your webroot (you're stuck with something like `public_html` or `htdocs`), you can set up Flarum without the `public` directory.
 
-If you intend to install Flarum using one of the archives, you can simply use the `no-public-dir` (Public Path = No) [archives](#installing-by-unpacking-an-archive) and skip the rest of this section. If you're installing via Composer, you'll need to follow the instructions below.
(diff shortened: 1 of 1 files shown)


Ser5:
Please note that the text may be incorrect as I didn't test it; also it makes `no-public-dir` archives obsolete.

luceos (commented):
Although I think your changes make a lot of sense, they assume a couple of things that might not be available to all users. 

Our public path installation instructions are kept simple for many reasons, for one we do not assume a separate directory is available outside the public path is available to install the "engine" into. 

Secondly your instructions seem to mix the engine path naming a few times, the fact that it's hard to understand what is what already makes it hard to read, I am not comfortable showing this level of complexity to new users with zero server knowledge.

I agree that the page needs some better instructions, clarifications why certain things are necessary for instance, and your PR is a good starting point to look at what areas to improve. But I don't think we should merge this as is.

I hope you understand, and are will to keep the PR open so we can improve our documentation based on it at a later point in time.

Thank you so much for your contribution and I wish you a fantastic 2025.

Ser5:
Oh hi there! Thank you for the review! When writing my version I already was sure that it will hardly match guidelines of your docs, terminology, etc. After all, a random dude out from nowhere can't magically understand all the ideas you have been working on for years. So yes, I fully agree that my PR shouldn't be bluntly merged, it's but a suggestion, that you can drag some ideas from.

> Thank you so much for your contribution and I wish you a fantastic 2025.
Wowowow, thanks a lot! I wish you the best too!

PR #535, opened 2026-06-11
Document the `flarum/audit` extension (1.x and 2.x)

imorland:
Adds documentation for the first-party **`flarum/audit`** extension, for both **1.x** and **2.x**.

Each version is split into two cross-linked pages, establishing a standard layout for documenting first-party extensions:

- **User / admin guide** (`extensions/audit`) — requirements, installation, permissions, browsing & search, the IP country indicator, anatomy of an entry, removing data, the full list of logged actions, and known limitations. Listed under a new **Bundled Extensions** sidebar category.
- **Developer guide** (`extend/audit`, "Integrating with Audit") — how any extension records its own actions via the public `Flarum\Audit\Extend\Audit` extender behind an optional `flarum/audit` dependency: event listeners, stateful `using()` integrations, the extender reference, and translating actions. Sits alongside the Realtime integration guide.

The 2.x docs additionally cover the new core/GDPR actions (group, developer token, settings reset, GDPR erasure/export) and recording the actor explicitly from a queued job.

Pairs with flarum/framework#4711.

### DIFF
--- a/docs/extend/audit.md
+++ b/docs/extend/audit.md
@@ -1,0 +1,125 @@
+# Integrating with Audit
+
+The [`flarum/audit` extension](../extensions/audit.md) records moderation and administration actions to a tamper-resistant log. It exposes a public extender, `Flarum\Audit\Extend\Audit`, so **any extension can record its own actions** without modifying Audit — the integration lives in the extension that owns the events.
+
+This is how every bundled extension (and, increasingly, the third-party ones) adds its audit coverage. For the catalogue of actions already recorded, see the [Audit extension page](../extensions/audit.md#logged-actions).
+
+:::info Optional dependency
+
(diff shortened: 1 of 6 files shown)


claudiushenrichs:
I'm a bit late to the game, but some suggestions:
- user guide: Use "audit log" consistently to refer to the log this extension keeps. Could be confused with the standard Flarum log that can be accessed via Log Viewer, so always referring to it by the full name should avoid ambiguity. Particularly important in the "Audit" introduction paragraphs.
- For the "Permissions" paragraph: Add a sentence on why the two permission levels exist, e.g. second level is to protect personal identifiable information to be exposed to a wider moderation team / for minimizing PII exposure to follow data protection regulation like GDPR

